PR #1, opened 2018-02-28
pester: generate XML reports

sodar:


sodar:
recheck no bug

sodar:
recheck no bug

jacek-izykowski:
recheck no bug

sodar:
recheck no bug

Jenkinsfile
```diff
@@ -38,6 +44,23 @@ pipeline {
...
+        stage('Prepare-Logs-Destination') {
+            agent { label 'master' }
+            steps {
+                script {
+                    if (env.ZUUL_UUID != null) {
+                        logsRemoteDir = "${logRootDir}/${env.ZUUL_UUID}"
+                        logsDestinationFile = "${logsRemoteDir}/log.txt.gz"
+                    } else {
+                        logsRemoteDir = "${logRootDir}/github/${env.JOB_NAME}/${env.BUILD_NUMBER}"
+                        logsDestinationFile = "${logsRemoteDir}/log.txt.gz"
```
m-kostrzewa: logsDestinationFile can be moved after the ifelse statement

Jenkinsfile
```diff
@@ -169,41 +192,38 @@ pipeline {
...
-                        // TODO correct flags for rsync
-                        sh "ssh ${LOG_SERVER_USER}@${LOG_SERVER} \"mkdir -p ${LOG_ROOT_DIR}/${BUILD_SPECIFIC_DIR}\""
-                        def tmpLogFilename = "clean_log.txt.gz"
-                        obtainLogFile(env.JOB_NAME, env.BUILD_ID, tmpLogFilename)
-                        sh "rsync $tmpLogFilename ${LOG_SERVER_USER}@${LOG_SERVER}:${REMOTE_DST_FILE}"
-                        deleteDir()
-                    }
+                    def tmpLogFilename = "clean_log.txt.gz"
+                    obtainLogFile(env.JOB_NAME, env.BUILD_ID, tmpLogFilename)
+                    sh "rsync $tmpLogFilename ${logServerUser}@${logServer}:${logsDestinationFile}"
```
m-kostrzewa: IMHO this whole `sh "rsync $tmpLogFilename ${logServerUser}@${logServer}:` should be abstracted into a function`pushToLogServer(filename, relative_filepath_on_the_fileserver)`

Then we can do magic like:

```
pushToLogServer(tmpLogFilename, "/log.txt.gz")
pushToLogServer("CIScripts/Test/testReport.xml", "testReports/testReport.xml")
```

Jenkinsfile
```diff
@@ -169,41 +192,38 @@ pipeline {
                             deleteDir()
                             unstash 'CIScripts'
                             powershell script: './CIScripts/Test.ps1'
+
+                            // TODO: you can do better
```
m-kostrzewa: No, sorry

m-kostrzewa: :fire:

Jenkinsfile
```diff
@@ -169,41 +192,38 @@ pipeline {
...
-    }
 
-    environment {
-        LOG_SERVER = "logs.opencontrail.org"
-        LOG_SERVER_USER = "zuul-win"
-        LOG_ROOT_DIR = "/var/www/logs/winci"
-        BUILD_SPECIFIC_DIR = "${ZUUL_UUID}"
-        JOB_SUBPATH = env.JOB_NAME.replaceAll("/", "/job/")
-        REMOTE_DST_FILE = "${LOG_ROOT_DIR}/${BUILD_SPECIFIC_DIR}/log.txt.gz"
+            post {
```
m-kostrzewa: This additional nested post stage seems unnecessary. Could you explain why is it here?

Jenkinsfile
```diff
@@ -38,6 +44,23 @@ pipeline {
             }
         }
 
+        stage('Prepare-Logs-Destination') {
+            agent { label 'master' }
+            steps {
+                script {
+                    if (env.ZUUL_UUID != null) {
+                        logsRemoteDir = "${logRootDir}/${env.ZUUL_UUID}"
```
m-kostrzewa: See, this is weird. If we were to go with the amazing `pushToLogServer(filename, relative_filepath_on_the_fileserver)`, then we need to tell it (inject) the `logsRemoteDir`. Which makes me think that it's actually a higher order function. 

We can basically construct our pushToLogServer function by injecting rootDir like this:
```
rootDir = "asdf/asdf/asdf/"
def pushLogs = { subdir -> rootDir + subdir }
print pushLogs ("myPrivateDir/asdf.txt")

// prints asdf/asdf/asdf/myPrivateDir/asdf.txt
```
Then, rootDir doesn't have to be global. However, pushLogs function would be global. But I think it's much better.

Jenkinsfile
```diff
@@ -38,6 +44,23 @@ pipeline {
             }
         }
 
+        stage('Prepare-Logs-Destination') {
+            agent { label 'master' }
```
m-kostrzewa: Is there any particular reason why this executed on master? Is it because env variables are only present there?

If that's the case, I would try to move this to `Preparation` stage somehow. If you think about it, `Preparation` stage, which basically stashes CI code, has no reason why it requires any ansible. Maybe it should be `master` there?

CIScripts/Test/TestRunner.ps1
```diff
@@ -46,29 +46,24 @@ function Invoke-TestScenarios {
...
             }; 
             Arguments=@()
         }
-
-        $Results = Invoke-Pester -PassThru -Script $PesterRunScript
-        $TotalFailedCount += $Results.FailedCount
     }
-    Write-Host "Num failed tests: $TotalFailedCount"
-    if ($TotalFailedCount -gt 0) {
+    $OutputXmlPath = "$PSScriptRoot\testReport.xml"
```
m-kostrzewa: `CIScripts/Test/testReport.xml` is a bad output path. We're mixing test results with scripts.

Let's put it somewhere in the root, outside of contrail-windows-ci directory altogether.

(this issue exists AFAIK for build artifacts and project repositories - they should never be stored/outputted inside CIScripts/ or contrail-windows-ci. But it's hard to decouple them all at the same time. So for now, let's not contribute to the mess further.)

How I see it for the nearest future:

```
contrail-windows-ci/{CIScripts,ansible,...}
bin/{...}
src/{controller,vrouter......}
testReport.xml <-- not sure exactly about this, but I don't think this should be outputted under contrail-windows-ci dir.
```

Jenkinsfile
```diff
@@ -169,41 +192,38 @@ pipeline {
                             deleteDir()
                             unstash 'CIScripts'
                             powershell script: './CIScripts/Test.ps1'
+
+                            // TODO: you can do better
+                            stash name: 'testReport', includes: 'CIScripts/Test/testReport.xml'
```
m-kostrzewa: See my comment about testReport output path. Here we can see that there's really something wrong. Why would test report file be contained within CIScripts dir?

Jenkinsfile
```diff
@@ -169,41 +192,38 @@ pipeline {
...
-        REMOTE_DST_FILE = "${LOG_ROOT_DIR}/${BUILD_SPECIFIC_DIR}/log.txt.gz"
+            post {
+                always {
+                    node('master') {
+                        script {
+                            try {
+                                unstash 'testReport'
+                                sh "rsync CIScripts/Test/testReport.xml ${logServerUser}@${logServer}:${logsRemoteDir}"
+                            } catch (err) {
+                                echo "No test reports available"
```
m-kostrzewa: Should we actually catch it or should we fail the build? Imho the latter.

Jenkinsfile
```diff
@@ -169,41 +192,38 @@ pipeline {
...
-                    if (env.ghprbPullId == null && env.JUNIPER_WINDOWSSTUBS == null) {
-                        // unstash "buildLogs"
-                        // TODO correct flags for rsync
-                        sh "ssh ${LOG_SERVER_USER}@${LOG_SERVER} \"mkdir -p ${LOG_ROOT_DIR}/${BUILD_SPECIFIC_DIR}\""
-                        def tmpLogFilename = "clean_log.txt.gz"
-                        obtainLogFile(env.JOB_NAME, env.BUILD_ID, tmpLogFilename)
-                        sh "rsync $tmpLogFilename ${LOG_SERVER_USER}@${LOG_SERVER}:${REMOTE_DST_FILE}"
-                        deleteDir()
-                    }
+                    def tmpLogFilename = "clean_log.txt.gz"
```
PawelMarchewkaCodiLime: Why not just ```logFilename```?

Jenkinsfile
```diff
@@ -10,6 +10,12 @@ def testEnvFolder
 
 def testbeds
 
+def logServer = "logs.opencontrail.org"
```
PawelMarchewkaCodiLime: Does jenkinsfile have support for structures or classes ? It would be good to group those variables if it is possible.

Jenkinsfile
m-kostrzewa: sry it should be something like
```
def pushLogs
if(true) { // simulate block
    rootDir = "asdf/asdf/asdf/"
    pushLogs = { subdir -> rootDir + subdir }
}
print pushLogs ("myPrivateDir/asdf.txt")
```

Jenkinsfile
```diff
@@ -38,6 +46,16 @@ pipeline {
             }
         }
 
+        stage('Preparation-Logs') {
+            agent { label 'master' }
+            steps {
+                script {
+                    logsDestination = decideLogsDestination()
+                    prepareLogsDestination(logsDestination)
```
m-kostrzewa: maybe move to post?

Jenkinsfile
```diff
@@ -169,43 +187,33 @@ pipeline {
                             deleteDir()
                             unstash 'CIScripts'
                             powershell script: './CIScripts/Test.ps1'
+                            stash name: 'testReport', includes: 'testReport.xml'
```
m-kostrzewa: try..finally?

Jenkinsfile
```diff
@@ -169,43 +187,33 @@ pipeline {
...
+                    } catch (err) {
+                        // unstash may throw an exception when pipeline failed before Test stage,
+                        // but we still need logs/metrics to be pushed
                     }
                 }
 
+                script {
+                    def logFilename = "log.txt.gz"
+                    obtainLogFile(env.JOB_NAME, env.BUILD_ID, logFilename)
+                    publishToLogServer(logFilename, logsDestination)
```
m-kostrzewa: There's some meh duplication here with multiple calls to publishToLogServer. not sure how to best get rid of it though

vars/prepareLogsDestination.groovy
```diff
@@ -0,0 +1,3 @@
+def call(String dest) {
+    sh "ssh ${env.LOG_SERVER_USER}@${env.LOG_SERVER} \"mkdir -p ${dest}\""
```
m-kostrzewa: hm, do we actually need to call this function explicitely, or should publishtoLogServer test if the directory exists and create it if it doesn't?

vars/pushToLogServer.groovy
```diff
@@ -0,0 +1,6 @@
+def logServerUser
+def logServer
+
+def call(String src, String destination) {
```
m-kostrzewa: seems like unused file?

vars/decideLogsDestination.groovy
```diff
@@ -0,0 +1,11 @@
+def call() {
+    def logsRemoteDir
+
+    if (env.ZUUL_UUID != null) {
```
m-kostrzewa: we use a hidden global variable here; it should be passed as param instead

Jenkinsfile
```diff
@@ -180,30 +184,25 @@ pipeline {
...
-                        sh "ssh ${LOG_SERVER_USER}@${LOG_SERVER} \"mkdir -p ${LOG_ROOT_DIR}/${BUILD_SPECIFIC_DIR}\""
-                        def tmpLogFilename = "clean_log.txt.gz"
-                        obtainLogFile(env.JOB_NAME, env.BUILD_ID, tmpLogFilename)
-                        sh "rsync $tmpLogFilename ${LOG_SERVER_USER}@${LOG_SERVER}:${REMOTE_DST_FILE}"
-                        deleteDir()
-                    }
+                    def logServer = [
+                        addr: env.LOG_SERVER,
+                        user: env.LOG_SERVER_USER,
+                        rootDir: env.LOG_ROOT_DIR
```
m-kostrzewa: In lines 184-186 those envs are defined. Maybe let's just use specific values here instead and remove lines 184-186?

CIScripts/Test/TestRunner.ps1
```diff
@@ -46,29 +47,24 @@ function Invoke-TestScenarios {
...
-        $Results = Invoke-Pester -PassThru -Script $PesterRunScript
-        $TotalFailedCount += $Results.FailedCount
     }
-    Write-Host "Num failed tests: $TotalFailedCount"
-    if ($TotalFailedCount -gt 0) {
+    $TestReportOutputPath = "$TestReportOutputDirectory\testReport.xml"
+    $Results = Invoke-Pester -PassThru -Script $PesterScripts `
+        -OutputFormat NUnitXml -OutputFile $TestReportOutputPath
+    Write-Host "Num failed tests: $($Results.FailedCount)"
+    Write-Host "Report written to $TestReportOutputPath"
```
jacek-izykowski: Period should be placed at the end of a sentence.
Please note: this is low priority issue.

jacek-izykowski: PM: Please add also information about number of passed tests.

Jenkinsfile
```diff
@@ -180,30 +184,25 @@ pipeline {
...
+                    ]
+                    def destDir = decideLogsDestination(logServer, env.ZUUL_UUID)
+
+                    unstash 'testReport'
```
jacek-izykowski: Complementary stash may never happen.
Especially when DONT_CREATE_TESTBEDS is set.
It may likely affect successful execution of a pipeline.

Jenkinsfile
```diff
@@ -180,30 +184,25 @@ pipeline {
...
+                    publishToLogServer(logServer, 'testReport.xml', destDir)
+
+                    def logFilename = 'log.txt.gz'
+                    obtainLogFile(env.JOB_NAME, env.BUILD_ID, logFilename)
+                    publishToLogServer(logServer, logFilename, destDir)
```
jacek-izykowski: What do you think about marking job as failed when there is no Jenkins log file? Such a situation should bring some attention to Jenkins' well being.

sodar:
Rebased on current `development`.

PR #2, opened 2018-02-28
Refactor Monitoring

jablonskim:


monitoring/tests/test_collectors.py
```diff
@@ -0,0 +1,238 @@
...
+class TestJenkinsGetBuildStatsEndpoint(unittest.TestCase):
+    def test_build_stats_endpoint_is_good(self):
+        url = JenkinsCollectorAdapter.get_build_stats_endpoint(build_url='http://localhost:8080/job/MyJob/1')
+        self.assertEqual(url, 'http://localhost:8080/job/MyJob/1/wfapi/describe')
+
+
+class TestJenkinsGetRawStats(unittest.TestCase):
+    def test_get_raw_stats_ok(self):
+        with requests_mock.mock() as m:
+            m.get('http://localhost:8080/job/MyJob/1/wfapi/describe', json={
```
m-kostrzewa: This is really minor, and Dariusz disagrees, but IMO we shouldn't use a "realistic" URL like `localhost` and port `8080` as test strings. It may be confusing. I would suggest something like `asdf:5678`. This would need to be changed throughout all files.

monitoring/tests/test_collectors.py
```diff
@@ -0,0 +1,238 @@
...
+            collector = JenkinsCollectorAdapter('MyJob', 'http://localhost:8080/job/MyJob/1')
+            stats = collector.get_raw_stats_from_jenkins()
+
+        self.assertIsNotNone(stats)
+        self.assertEqual(stats['id'], 1)
+        self.assertEqual(stats['endTimeMillis'], 1234)
+        self.assertEqual(stats['status'], 'SUCCESS')
+        self.assertEqual(stats['durationMillis'], 1000)
+
+    def test_get_raw_stats_error(self):
```
m-kostrzewa: Pedantic, but we should probably name it `test_get_raw_stats_invalid_build_id` or something.

monitoring/tests/test_collectors.py
```diff
@@ -0,0 +1,238 @@
+#!/usr/bin/env python3
```
m-kostrzewa: We had a discussion about all these tests. 
We think they kinda smell right now. They smell of Rigidity.

> Rigidity. The software is difficult to change. A small change causes a cascade of subsequent changes.

I think the reason for that is that we are testing what are private methods of JenkinsCollectorAdapter. We should only test via collect() method - the other ones should be private (maybe add `_` before their names).

There's also quite a lot of code duplication in tests (repeated asserts).

Public iface of JenkinsCollectorAdapter is that it consumes Jenkins API and produces *Stats object. We should test the happy case, some error cases that might happen and see if an exception is raised.

monitoring/collectors/jenkins_collector_adapter.py
```diff
@@ -0,0 +1,59 @@
...
+        self.job_name = job_name
+        self.url = build_url
+
+
+    def collect(self):
+        resp = self.get_raw_stats_from_jenkins()
+        return self.convert_raw_stats_to_build_stats(resp)
+
+
+    def get_raw_stats_from_jenkins(self):
```
m-kostrzewa: See our comments for this class tests, but get_raw_stats_from_jenkins and convert_raw_stats_to_build_stats should probably be private and not tested directly.

monitoring/collectors/jenkins_collector_adapter.py
```diff
@@ -0,0 +1,59 @@
...
+            status = raw_stats['status'],
+            duration_millis = raw_stats['durationMillis'],
+            stages = stages_stats,
+        )
+
+        return build_stats
+
+
+    @classmethod
+    def get_build_stats_endpoint(cls, build_url):
```
m-kostrzewa: get_build_stats_endpoint should probably be private and not tested directly as well. 

If we want to test it, let's write tests that instantiate JenkinsCollectorAdapter with different URLs and see if collect() triggers a request received by mocked API endpoint at the specified URLs, e.g.:

```
j = JenkinsCollectorAdapter(build_url="wp.pl/myJob/.../x", job_name="myJob")
with request_mock_assert_receives("wp.pl/myJob/.../x"):
   j.collect()
```

m-kostrzewa: also, then we don't see a reason for this being a `@classmethod `

monitoring/publishers/mysql_publisher_adapter.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+    def publish(self, build_stats):
+        build = Build.from_build_stats(build_stats)
+        self.session.add(build)
+        self.session.commit()
+
+
+    @classmethod
+    def get_connection_string(cls, host, username, password, database):
```
m-kostrzewa: get_connection_string - same comment as for get_build_stats_endpoint. I think we should make it 1. private and 2. tested via Publisher API only.

monitoring/publishers/mysql_publisher_adapter.py
```diff
@@ -0,0 +1,30 @@
...
+        return 'mysql://{}:{}@{}/{}'.format(username, password, host, database)
+
+
+    @classmethod
+    def get_database_session(cls, connection_string):
```
m-kostrzewa: We think that this method shouldn't be static.

monitoring/tests/test_publishers.py
```diff
@@ -0,0 +1,110 @@
...
+            job_name = 'Test',
+            build_id = 1234,
+            build_url = 'http://localhost:8080/job/MyJob/1',
+            finished_at_secs = 5678,
+            status = 'SUCCESS',
+            duration_millis = 4321,
+            stages = [],
+        )
+
+        with patch.object(MySQLPublisherAdapter, 'get_database_session', return_value=self.session):
```
m-kostrzewa: We had some discussion here. get_database_session is an implementation detail of MySQLPublisherAdapter. It should be private.

But then, we mock a private method here (and in other tests). Which kind of smells and suggests that there's another class lurking somewhere (or something else).

Because get_database_session can be parametrized by connection string, maybe we should inject connection_string to MySQLPublisherAdapter and rename it to DatabasePublisherAdapter? A more involved solution would be to implement a MySQLSessionHandler class with get_connection_string and get_database_session method but it seems too complicated for now, so let's not do it.

In any case, _get_database_session should be private, and mocking a private method smells.

m-kostrzewa: Actually, why not have an interface called `DatabaseSession`. It would have a method `get_database_session`. 

`DatabaseSession` would be injected to `DatabasePublisherAdapter`.

We would have two implementations: `MySQLSession` and `StubSQLiteSession`. We could use the former in production code and the latter in tests.

Then in the test set up we would have something like:
```
create_engine("sqlite://")
self.session = StubSQLiteSession("sqlite://")
...
```

monitoring/tests/test_publishers.py
```diff
@@ -0,0 +1,110 @@
...
+            publisher = MySQLPublisherAdapter('', '', '', '')
+            publisher.publish(build_stats)
+
+        build_count = self.session.query(Build).count()
+        self.assertEqual(build_count, 1)
+        build = self.session.query(Build).one()
+        self.assertEqual(len(build.stages), 2)
+        self.assertIsNotNone(build.stages[0])
+        self.assertIsNotNone(build.stages[1])
```
m-kostrzewa: Unit tests should ideally be like 5 lines long.

Maybe we should do some helper functions, like `assert_stages_equal(stage1_stats, build.stages[0])`.

Also, maybe we should move test objects, like `stage2_stats` into set_up or just as a class field.

The goal here is to make the actual test body as short and clear as possible.

This goes for all other such "long" tests in this PR. 

Then, this test would look like:
```
build_stats_with_two_stages = (..., stages=[...])

...

def test_publish_build_stats_with_stages(self):
    mock_something_according_to_other_PR_comment()
    publisher = MySQLPublisherAdapter('', '', '', '')
    publisher.publish(build_stats_with_two_stages)
    assert_has_two_stages(build)
    assert_stages_equal(stage1_stats, build.stages[0])
    assert_stages_equal(stage2_stats, build.stages[1])
```

monitoring/finished_build_stats_publisher.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class FinishedBuildStatsPublisher(object):
+
+    def __init__(self, collector, publisher):
+        self.collector = collector
+        self.publisher = publisher
+
+    def collect_and_publish(self, delay_ms=1000):
+        while True:
```
m-kostrzewa: This really needs a timeout. Just sayin'. ;)

monitoring/finished_build_stats_publisher.py
```diff
@@ -0,0 +1,18 @@
...
+            stats = self.collector.collect()
+            if stats.status != 'IN_PROGRESS':
```
m-kostrzewa: We think that instead of this condition, stats object should have a method `is_finished()`, which should be used here.

monitoring/finished_build_stats_publisher.py
```diff
@@ -0,0 +1,18 @@
...
+                break
+
+            time.sleep(delay_ms / 1000.0)
```
m-kostrzewa: And then we need a good timeout error message exception here in case it happens.

monitoring/finished_build_stats_publisher.py
m-kostrzewa: Maybe something like 30 secs?

monitoring/provision_mysql_database.py
```diff
@@ -0,0 +1,29 @@
...
+
+def provision_database(connection_string, model):
+    engine = create_engine(connection_string, echo=True)
+    model.metadata.create_all(engine)
+
+
+def main():
+    args = parse_args()
+
+    connection_string = MySQLPublisherAdapter.get_connection_string(host=args.mysql_host,
```
m-kostrzewa: This code will probably have to change due to other CR changes.

monitoring/tests/test_finished_build_stats_publisher.py
```diff
@@ -0,0 +1,64 @@
...
+        collector.collect = MagicMock(side_effect=TestFinishedBuildStatsPublisher.CollectSideEffect())
+
+        publisher = IPublisher()
+        publisher.publish = MagicMock()
+
+        stats_publisher = FinishedBuildStatsPublisher(collector, publisher)
+        stats_publisher.collect_and_publish(delay_ms=0)
+
+        self.assertEqual(collector.collect.call_count, 3)
+        publisher.publish.assert_called_once_with(success_build_stats)
```
m-kostrzewa: These two tests share a lot of boilerplate code. We should probably refactor it somehow.

monitoring/tests/test_finished_build_stats_publisher.py
```diff
@@ -0,0 +1,64 @@
...
+import unittest
+from unittest.mock import MagicMock
+from finished_build_stats_publisher import FinishedBuildStatsPublisher
+from collector import ICollector
+from publisher import IPublisher
+from stats import BuildStats
+
+
+class TestFinishedBuildStatsPublisher(unittest.TestCase):
+    class CollectSideEffect(object):
```
m-kostrzewa: Maybe definition of this class should be inside `test_with_retries`.

Maybe it should be a generator, like:
```
>>> def c():
...     for i in range(3):
...         yield i
...     return
```

m-kostrzewa:
reviewed by DS +

jablonskim:
The [line](https://github.com/Juniper/contrail-windows-ci/pull/2/files#diff-9e3ee9dd222aad79a3d15b76b4da1b48R62) has been taken from some sqlalchemy example. It provides reversible relationship between `Stage` and `Build`.

monitoring/tests/test_finished_build_stats_publisher.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_finished(self):
+        self.collector.collect = MagicMock(return_value=self.build_stats_success)
+
+        self.stats_publisher.collect_and_publish()
+
+        self.collector.collect.assert_called_once_with()
+        self.publisher.publish.assert_called_once_with(self.build_stats_success)
+
+    def test_with_retries(self):
+        def generator():
```
m-kostrzewa: Maybe rename to counter?

monitoring/stats.py
```diff
@@ -1,42 +1,21 @@
...
+    'stages'
+])
 
-    if 'stages' in resp.keys():
-        for resp_stage in resp['stages']:
-            stage = Stage(name=resp_stage['name'],
-                          status=resp_stage['status'],
-                          duration_millis=resp_stage['durationMillis'])
-            build.stages.append(stage)
+BuildStats.is_finished = lambda self: self.status != 'IN_PROGRESS'
```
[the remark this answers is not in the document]
jablonskim: It will be easy to do using inheritance (like in this [example](https://stackoverflow.com/a/3223262)).

monitoring/tests/test_stats.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+from tests.common import get_test_build_stats_with_status
+
+
+class TestIsFinished(unittest.TestCase):
```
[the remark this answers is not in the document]
jablonskim: I am testing `is_finished` method here. The `get_test_build_stats_with_status` function is used only to create `stats` with specific `status`.

monitoring/tests/test_finished_build_stats_publisher.py
```diff
@@ -0,0 +1,52 @@
...
+
+        self.collector.collect = MagicMock(side_effect=generator())
+
+        self.stats_publisher.collect_and_publish(delay_ms=0)
+
+        self.assertEqual(self.collector.collect.call_count, 3)
+        self.publisher.publish.assert_called_once_with(self.build_stats_success)
+
+    @timeout_decorator.timeout(2)
+    def test_timeout(self):
```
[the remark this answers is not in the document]
jablonskim: I decided to use this decorator only for testing purposes. Personally I prefer the `max_retries` and `MaxRetriesExceededError` approach in production code.

monitoring/collectors/jenkins_collector_adapter.py
```diff
@@ -0,0 +1,58 @@
...
+        timestamp = int(raw_stats['endTimeMillis'] / 1000)
+
+        stages_stats = []
+
+        if 'stages' in raw_stats.keys():
+            stages_stats = [StageStats(
+                name = raw_stage['name'],
+                status = raw_stage['status'],
+                duration_millis = raw_stage['durationMillis']
+            ) for raw_stage in raw_stats['stages']]
```
m-kostrzewa: This is pretty minimal and pedantic, but I confused raw_stage with raw_stats and I lost 5 seconds of my life figuring it out. Maybe rename raw_stage to x or stage or something which doesn't have levenshtein distance of 2.

monitoring/finished_build_stats_publisher.py
```diff
@@ -0,0 +1,24 @@
...
+class FinishedBuildStatsPublisher(object):
+
+    def __init__(self, collector, publisher):
+        self.collector = collector
+        self.publisher = publisher
+
+    def collect_and_publish(self, delay_ms=1000, max_retries=10):
+        while max_retries != 0:
+            stats = self.collector.collect()
+            if stats.is_finished():
```
m-kostrzewa: Also nitpicky, but maybe it should be .is_build_finished? Since stats can't be finished.

monitoring/tests/test_finished_build_stats_publisher.py
```diff
@@ -0,0 +1,52 @@
...
+class TestFinishedBuildStatsPublisher(unittest.TestCase):
+    def setUp(self):
+        self.collector = MagicMock()
+        self.publisher = MagicMock()
+        self.publisher.publish = MagicMock()
+        self.stats_publisher = FinishedBuildStatsPublisher(self.collector, self.publisher)
+        self.build_stats_success = get_test_build_stats_with_status('SUCCESS')
+        self.build_stats_in_progress = get_test_build_stats_with_status('IN_PROGRESS')
+
+    def test_finished(self):
```
m-kostrzewa: test_finished name is weak and will not survive the winter. Maybe something like test_collecting_already_finished_build would be better?

monitoring/tests/test_publishers.py
```diff
@@ -0,0 +1,72 @@
...
+        build_stats = get_test_build_stats([TEST_STAGE1_STATS, TEST_STAGE2_STATS])
+        self.publisher.publish(build_stats)
+
+        build_count = self.session.query(Build).count()
+        self.assertEqual(build_count, 1)
+
+        build = self.session.query(Build).one()
+        self.assertEqual(len(build.stages), 2)
+        self.assertIsNotNone(build.stages[0])
+        self.assertIsNotNone(build.stages[1])
```
m-kostrzewa: These assertIsNotNone seem redundant, next asserts will check them basically.

PR #3, opened 2018-02-28
Add an ICMP utils-based test

krdln:


krdln:
@m-kostrzewa 
> If anything, we should generate a config file in CI that would be passed to tests.

Yeah, currently I'm just trying to make the test pass, I'll try to clean up afterwards, so the envs in the file are only a temporary solution. Generating the file at runtime seems like a reasonable idea.

> The tests should "query" env (which is dependent on CI) in any way.

You mean shouldn't, right?

krdln:
TODO:
* All the TODOs introduced in commits
* Fix for @m-kostrzewa comment above (generate the `GetTestConfiguration` file in CI, rather than use envs)
* Merge https://github.com/Juniper/contrail-windows-ci/pull/8 (and update template env)
* Merge https://github.com/Juniper/contrail-windows-ci/pull/6 (and update template env)
* Port TCP test to pester

michal-clapinski:
Since #6 won't build IIS server, it should be probably done in this PR. @katrybacka

krdln:
recheck no bug

krdln:
It seems everything works now [except the failing agent MSI installer test](http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/861/testReport.html).

[Error message](http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/861/log.txt.gz#_2018-03-15_20_50_24_485000)

krdln:
(the latest commit is already in PR #33 but I'll leave it also on this branch, so this PR has more chance to get through CI)

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -135,21 +108,26 @@ function Enable-DockerDriver {
...
-            Param ($Cfg, $ControllerIP, $Tenant, $Adapter)
-
-            $Env:OS_USERNAME = $Cfg.Username
-            $Env:OS_PASSWORD = $Cfg.Password
-            $Env:OS_AUTH_URL = $Cfg.AuthUrl
-            $Env:OS_TENANT_NAME = $Tenant
-
-            & "C:\Program Files\Juniper Networks\contrail-windows-docker.exe" -forceAsInteractive -controllerIP $ControllerIP -adapter "$Adapter" -vswitchName "Layered <adapter>" -logLevel "Debug"
-        } -ArgumentList $Configuration, $ControllerIP, $TenantName, $AdapterName | Out-Null
+            Param ($OpenStack, $ControllerIP, $Adapter)
```
jablonskim: Maybe we should exclude this rule which checks variables if it includes credentials in PSScriptAnalyzer instead of renaming `OSCreds` to `Openstack`

krdln: I've tried to suppress the message here, but without success. We've agreed with @katrybacka that we should just exclude this rule globally and from what I remember, she's planning to do it.

Anyway, I'm going to refactor this function in the next pull request, so I don't think it makes sense to change it here. Do you agree?

katrybacka: Actually i added this comment, we were doing this review together without Mateusz. For me it is fine to change it in other pr, i was planning to do the excluding of this mentioned rule anyway.

PR #6, opened 2018-03-01
Download microsoft/iis

michal-clapinski:


ansible/roles/testbed/tasks/main.yml
```diff
@@ -52,16 +52,30 @@
...
+- name: Copy IIS TCP test Dockerfile
+  win_copy:
+    src: "{{dockerfile_dir}}"
+    dest: C:\
+
+# docker_image module doesn't work on Windows
+- name: Build IIS TCP test Docker image
+  register: output
+  win_shell: |
+    docker build -t iis-tcptest "C:\\{{dockerfile_dir}}"
```
m-kostrzewa: Hmm. (rambling mode on)

I'm not sure I like this long-term. The reason is that we kinda couple the tests into the system that provides the environment. Tests shouldn't depend on the environment - it should be the other way around.

Ideal scenario is that someone clones the tests, edits the test configuration file and just runs Invoke-Pester in the root. This is not only easy to use by the devs, but is also decoupling modules that shouldn't really be coupled.

I guess the problem is that when `docker run ...` is invoked inside the test script, it doesn't build the image. `docker run` will pull images from dockerhub if need be, or use the images stored locally. Maybe a better scenario would be for the test itself to build the image if it's not present?

**tl;dr**

I guess I'm thinking that `docker build` should be a part of set up of the TCP test itself. We could prebuild it in the testbed image to speed up the process but we shouldn't rely on it.

So IMHO: this ansible change should stay but this step should also be added into the TCP test setup.

krdln: How does the `docker build` work after stopping docker service? I hope it doesn't auto-start here, because we need docker to be disabled when creating a template.

PR #7, opened 2018-03-02
grafana: initial version of dashboards in YAML

sodar:
- `grafyaml` does not support `table` as panel type as of yet

### DIFF
--- a/monitoring/README.md
+++ b/monitoring/README.md
@@ -36,3 +36,36 @@
     ```
     $ python3 test_stats.py
     ```
+
+### Grafana dashboards
+
+Grafana dashboards are described in YAML files.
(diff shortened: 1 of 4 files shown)


krdln:
I wonder if it's possible to get rid of this duplication here. Would be nice to do it, but I don't have any clear ideas on how.

Also, I see that you didn't add the `$Status` and `$Job` variables. I felt like quickly filtering by failed jobs was a nice feature. Also, if we used `$Job` variable, we could have just one dashboard, and access it by different urls based on `prod`/`devel`.

sodar:
@krdln I agree, we need to deduplicate the code in future. There is some templating support in `grafyaml`, but I did not figure it out yet. Same with `$Status` and `$Job` variable.

krdln:
You're using `AND status in ($Status)` only in one of the graphs. Is this on purpose?

sodar:
I forgot about the rest. I will add them.

PR #8, opened 2018-03-02
Pull microsoft docker images in testbed template

krdln:
The goal is for testbeds not to have network connection, so we should do it in template.


krdln:
Unfortunately [it took 22 minutes to pull the images](http://10.84.12.26:8080/job/WinContrail-infra/job/templates/job/create-testbed-template/25/console).

I think that to speed up this process when testing template creation, one can existing template image as a `VM_TEMPLATE` (base OS image) parameter.

PR #17, opened 2018-03-06
pester: generate report in HTML

sodar:
Blocked by: https://github.com/Juniper/contrail-windows-ci/pull/1

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -188,6 +188,19 @@ pipeline {
 
     post {
         always {
+            node('tester') {
+                deleteDir()
+                unstash 'CIScripts'
+                script {
+                    try {
+                        unstash 'testReport'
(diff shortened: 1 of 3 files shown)


sodar:
recheck no bug

sodar:
recheck no bug

Jenkinsfile
```diff
@@ -188,6 +188,19 @@ pipeline {
 
     post {
         always {
+            node('tester') {
+                deleteDir()
+                unstash 'CIScripts'
+                script {
+                    try {
+                        unstash 'testReport'
+                        powershell script: './CIScripts/GenerateTestReport.ps1'
```
jacek-izykowski: What do you think about passing report file name(s) as a parameter(s) to GenerateTestReport?

sodar: I think it might be a good addition. We'll change when reports from othere testing frameworks will occur.

PR #27, opened 2018-03-13
Test stubs for all test scenarios.

m-kostrzewa:
* Added test stubs with -Pending flag for all test cases *- to reviewer: I left some comments in the code, please help resolve them if you can, otherwise I would leave them for the implementor*
* Created some directories and moved stuff around - it's easiest to see not in diff, but when just looking at the source tree directly

Also FYI: I want to move the whole Tests/ directory to root in the future.

EXAMPLE as of 3f02812: http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/814/testReport.html


CIScripts/Test/Tests/Functional/DockerDriverNetworking.Tests.ps1
```diff
@@ -0,0 +1,52 @@
+Describe "Docker driver networking" {
+    Context "conflicting local IPs" {
+        It "can create containers for different tenants with same IP" -Pending {
```
krdln: Is there something like "Skipped" in pester? (I guess not, so we're using pending as skipped)

CIScripts/Test/Tests/Functional/WindowsLinuxInterop.Tests.ps1
```diff
@@ -0,0 +1,13 @@
+Describe "Windows Linux interoperation tests" {
+    It "ICMP connectivity works" -Pending {
```
krdln: [opinion] We think it should be moved to Protocol tests.

CIScripts/Test/Tests/Integration/ForwardingExtension.Tests.ps1
```diff
@@ -0,0 +1,5 @@
+Describe "Forwarding extension" {
+    It "enabling/disabling multiple consective times doesn't crash the system" -Pending {
```
krdln: [refinement] Typo – consecutive

CIScripts/Test/Tests/Integration/SNAT.Tests.ps1
```diff
@@ -0,0 +1,13 @@
+Describe "SNAT manual setup" {
```
krdln: [opinion] We're not sure what this test does, but perhaps it should be moved to Functional?

CIScripts/Test/Tests/Protocol/TunnellingWithAgent.Tests.ps1
```diff
@@ -0,0 +1,25 @@
+Describe "Tunnelling with Agent tests" {
+    Context "ICMP" {
+        It "works over MPLSoGRE" -Pending {
```
krdln: [refinement] We think the grouping should be reversed – MPLSo{GRE/UDP} as `Context`s and ICMP/TCP/UDP as `It` tests

CIScripts/Test/Tests/Reliability/ForwardingExtLongLeak.Tests.ps1
```diff
@@ -0,0 +1,5 @@
+Describe "Forwarding extension long leak test" {
+    It "doesn't crash the system when idle for a long time" {
```
krdln: [bug] Missing `-Pending` (?)

krdln: [bug] Missing `-Pending`

CIScripts/Test/Tests/Functional/DockerDriverNetworking.Tests.ps1
krdln: We see that tests without `Should` are already marked as "Inconclusive". What's the reason for `-Pending` then?

m-kostrzewa: Skipped is for tests that are implemented, but are to be skipped, e.g. because they are too long or we don't want to run them in CI.

I don't like the "inconclusive as default" behaviour of pester. It should be "error by default".

-Pending is there to signify that I haven't forgotten to write the test.

CIScripts/Test/Tests/Integration/SNAT.Tests.ps1
m-kostrzewa: I removed Functional entirely for now

m-kostrzewa:
> [refinement] As it's not obvious which test should go where, we think there should be a readme added to the Tests directory that describes what makes a test an "integration test" etc. Alternatively, separate readmies in Functional/Protocol etc. directories.

Removed Functional for now. Distinction between Protocol/Integration seems more obvious. Do you think there's still a need for a readme now?

PR #33, opened 2018-03-16
Fix agent MSI installer tests flakiness

krdln:
Make sure there are no leftover handles to Agent service:
* Don't pass the Service object across the session boundary
* Run GC after/before agent installation

Also, lower the timeout in MSI test as it's more robust now.

cc @katrybacka @m-kostrzewa


CIScripts/Test/Utils/ComponentsInstallation.ps1
```diff
@@ -33,6 +38,12 @@ function Uninstall-Agent {
 
     Write-Host "Uninstalling Agent"
     Invoke-MsiExec -Uninstall -Session $Session -Path "C:\Artifacts\contrail-vrouter-agent.msi"
+
+    # Get rid of all leftover handles to the Agent service
+    Invoke-Command -Session $Session -ScriptBlock {
```
m-kostrzewa: Maybe this should be part of Invoke-MsiExec?

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -215,13 +215,15 @@ function Disable-AgentService {
...
-        return $null
+
+    Invoke-Command -Session $Session -ScriptBlock {
+        $Service = Get-Service "ContrailAgent" -ErrorAction SilentlyContinue
+
+        if ($Service -and $Service.Status) {
+            return $Service.Status.ToString()
+        } else {
+            return $null
+        }
```
m-kostrzewa: > Don't pass the Service object across the session boundary

if we do this fix, do we actually need changes to MsiExec?

krdln: I don't know actually. I'd say a risk of Service handle not being GC-ed exists also if we don't pass it through the boundary.

CIScripts/Test/Utils/ComponentsInstallation.ps1
krdln: Only agent registers a service so this seems more Agent specific. But I'm open to doing it other way.

CIScripts/Test/TestConfigurationUtils.ps1
krdln: From my very imprecise testing it seems that not passing Service across the session boundary indeed lowers the average time of the test, but running GC lowers it even more.

CIScripts/Test/Utils/ComponentsInstallation.ps1
PawelMarchewkaCodiLime: Actually DockerDriver is also registered as a service, moreover someday it will be run as a service, so I would suggest to move it inside Invoke-MsiExec

m-kostrzewa: (there were also plans to have node-manager service and forwarding-extension service, maybe orchestrator in the future as well. So this would future proof them)

CIScripts/Test/TestConfigurationUtils.ps1
m-kostrzewa: OK let's keep it then I guess. Nice find though

krdln:
recheck no bug

PR #34, opened 2018-03-16
Run vtest case-by-case and disable outdated tests

krdln:
Stop using the `all_tests_run.ps1` script, but run the tests directly from Pester.

Also, mark the tests that use `NH_FLAG_ENCAP_L2` in `vr_nexthop_req` as *inconclusive* (temporarily)


krdln:
[Example report](http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/887/testReport.html)

m-kostrzewa:
@krdln pls can you choose different reviewer?

krdln:
@m-kostrzewa Blacklisting is no longer necessary and running tests test-by-case is low priority. I think we should just revisit this in the future.

PR #36, opened 2018-03-19
testenv: move testenv block configuration to ansible vars

sodar:


sodar:
recheck no bug

vars/demoEnvPipeline.groovy
```diff
@@ -20,23 +18,23 @@ def call(playbookToRun) {
                 steps {
```
krdln: Is it worth splitting in two stages at all, if the first stage only sets some variables?

ansible/group_vars/all/vars
```diff
@@ -0,0 +1,23 @@
...
+provisioning:
+  name: "{{ testenv_name }}"
+  folder: "{{ testenv_vmware_folder }}"
+  portgroup_mgmt: "{{ testenv_mgmt_network }}"
+  portgroup_contrail: "{{ testenv_data_network }}"
+  vms:
+    controller:
+      template: "{{ testenv_controller_vmware_template }}"
+      nodes:
+      - name: "{{ testenv_name[:7] }}-controller"
```
krdln: Perhaps it'd be better to just error early if `testenv_name` is longer than 7?

krdln:
I still don't get why the PR is called "prepend hostname with machine type", could you explain?

ansible/group_vars/all/vars
```diff
@@ -0,0 +1,23 @@
+# Default test env templates
+testenv_testbed_vmware_template: Template-testbed-201803050718
+testenv_controller_vmware_template: Template-CentOS-7.4
+
+# Testenv block definition
+provisioning:
```
krdln: `provisioning` seems a little bit ambiguous, perhaps use the comment as variable name?

sodar: `testenv_block`?

vars/demoEnvPipeline.groovy
sodar: Not really, I will merge these stages.

ansible/group_vars/all/vars
sodar: Trigger an error in Jenkinsfile or add some logic to `vmware-deploy-testenv.yml` playbook?

PawelMarchewkaCodiLime: [Refinement] @sodar I strongly agree with @krdln, job should fail if ```testenv_name ``` is longer then 7 characters.

krdln: If the error is in ansible but before anything side-effectful happens, I'm ok with that. I'd also check the machine name length instead of just testenv name, in case we change the naming scheme.

It'd be nice to also add a comment to param in deploy test and demo env jobs.

ansible/group_vars/all/vars
krdln: Sgtm

krdln:
@sodar Can you change the PR title then? :)

Jenkinsfile
```diff
@@ -117,35 +113,47 @@ pipeline {
             steps {
                 script {
                     lock(label: 'testenv_pool', quantity: 1) {
-                        testNetwork = getLockedNetworkName()
-                        vmwareConfig = getVMwareConfig()
-                        testEnvName = getTestEnvName(testNetwork)
-                        testEnvFolder = env.VC_FOLDER
+                        // 'Cleanup' stage
+                        node(label: 'ansible') {
```
krdln: This comment is unrelated to this block.

Also, this code could be run outside the `node` block. (if not, please explain why it needs to be run inside `node`)

ansible/group_vars/all/vars
```diff
@@ -0,0 +1,23 @@
+# Default test env templates
+testenv_testbed_vmware_template: Template-testbed-201803050718
+testenv_controller_vmware_template: Template-CentOS-7.4
```
krdln: Do we need these defaults here?

krdln:
Have you checked that deploy demo env works after this refactoring?

sodar:
@krdln Yup:
- deploying: http://10.84.12.26:8080/job/WinContrail-infra/job/deploy-dev-env/28/
- destroying: http://10.84.12.26:8080/job/WinContrail-infra/job/destroy-env/76
- name too long: http://10.84.12.26:8080/job/WinContrail-infra/job/deploy-dev-env/30

Jenkinsfile
sodar: It needs to be in `node`, because script uses `env.WORKSPACE` which is only defined when job is running on a given node.

But yes, comment can be deleted.

ansible/group_vars/all/vars
sodar: I was thinking about these as a substitute to Jenkins parameters/contents. I think they can stay for now and in the future we'll move jenkins variables to vars/vault.

PR #40, opened 2018-03-20
Exclude PSAvoidUsingPlainTextForPassword rule from PSScriptAnalyzer

katrybacka:
Ecxlude this rule due to bug with 0SCreds variable in testenv (SuppressMessage doesn't work)


m-kostrzewa:
In this case, I think we need to grep all occurrences of manual suppression and remove them

krdln:
recheck no bug

PR #43, opened 2018-03-20
Don't try to unstash testReport since it may not exist

PawelMarchewkaCodiLime:


PawelMarchewkaCodiLime:
It should work on production http://10.84.12.26:8080/job/WinContrail/job/WinCI-pm-test/1/console

PR #45, opened 2018-03-20
Update utils-based test to use new flags

krdln:
and re-enable utils-based test


CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1
```diff
@@ -71,9 +71,22 @@ Describe "Single compute node protocol tests with utils" {
                 vif.exe --add $Using:Container1NetInfo.IfName --mac $Using:Container1NetInfo.MACAddress --vrf 1 --type virtual
                 vif.exe --add $Using:Container2NetInfo.IfName --mac $Using:Container2NetInfo.MACAddress --vrf 1 --type virtual
 
-                nh.exe --create 1 --vrf 1 --type 2 --el2 --oif $Using:Container1NetInfo.IfIndex
-                nh.exe --create 2 --vrf 1 --type 2 --el2 --oif $Using:Container2NetInfo.IfIndex
-                nh.exe --create 3 --vrf 1 --type 6 --cen --cni 1 --cni 2
+                try {
```
katrybacka: What do you think about re-enabling Extension building in CI and then we can merge your change which support only new utils, or update utils in Artifacts to new ones instead of using try-catch block here and supporting both versions?

krdln: I though about doing that, but that way, our Ci would be unusable from the moment of enabling extension to the moment of merging this pr. I prefer the two-stage way as it won't cause any downtime.

krdln:
Let's do #46 first, and then enable extension.

krdln:
recheck no bug

krdln:
recheck no bug

PR #47, opened 2018-03-21
test-report: catch unstashing errors

sodar:


Jenkinsfile
```diff
@@ -189,14 +189,16 @@ pipeline {
...
-                //         unstash 'testReport'
-                //         powershell script: './CIScripts/GenerateTestReport.ps1'
-                //     } finally {
-                //         stash name: 'testReport', includes: '*.xml,*.html', allowEmpty: true
-                //     }
-                // }
+                script {
+                    try {
+                        unstash 'testReport'
+                        powershell script: './CIScripts/GenerateTestReport.ps1'
```
krdln: Can we modify it so errors in `GenerateTestReport` are not treated as no test report? Perhaps
```groovy
def hasReports
try { unstash 'testReport'; hasReports = true }
catch (Exception _) { hasReports = false }
if (hasReports) { ... }
```

sodar:
Tested on pipeline cloned from production pipeline: http://10.84.12.26:8080/job/WinContrail/job/WinCi-ds-test/2

Jenkinsfile
sodar: `GenerateTestReport` should not fail if there are no test reports. Script checks whether each (currently 1) XML exist.

krdln: So I guess it could be moved to `finally` block?

sodar: Yes. I will move it there.

sodar:
recheck no bug

PR #48, opened 2018-03-21
Don't use COMPONENTS_TO_BUILD env variable from jenkins

PawelMarchewkaCodiLime:


PawelMarchewkaCodiLime:
Here is job proving that COMPONENTS_TO_BUILD in Jenkins is now ignored:
http://10.84.12.26:8080/job/WinContrail/job/WinCI-pm-test/8/console

krdln:
Let's wait before removing parameters until it's merged to production.

krdln:
Removed paramters from `-devel` `-cw-devel` `contrail-win-ci-gh` and `stubs-gh-devel` jobs. Production stays unchanged for now.

PR #50, opened 2018-03-22
Added docker image building

katrybacka:


katrybacka:
recheck no bug

katrybacka:
recheck no bug

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.Tests.ps1
```diff
@@ -10,4 +10,10 @@ Describe "Initialize-IisTcpTestDockerImage" {
     It "Builds iis-tcptest image" {
         { Initialize-IisTcpTestDockerImage -Session $Session } | Should Not Throw
     }
+    
+    It "iis-tcptest image appear in docker inspect" {
+        Invoke-Command -Session $Session {
+            { docker inspect iis-tcptest } | Should Not BeNullOrEmpty
+        }
```
krdln: * Should should be outside `Invoke-Command`
* Please remove `{}` braces before `| Should` – they are only needed in `Should [Not] Throw`.
* This testcase doesn't actually make sure that the build has been done before – we *shouldn't* rely on ordering of `It` clauses. Possible solutions:
    * Merge these two `It`s
    * Convert the first `It` to `BeforeEach`

jacek-izykowski: We suggest the following procedure:
1. Remove existing image from docker.
2. Build fresh image.
3. Verify that image has been created.

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.ps1
```diff
@@ -0,0 +1,18 @@
+. $PSScriptRoot\..\..\Common\Aliases.ps1
+$DockerfilePath = "$PSScriptRoot\..\..\DockerImages\iis-tcptest\Dockerfile"
+function Initialize-IisTcpTestDockerImage  {
+    Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session)
+
+    $DockerImagesDir = "C:\DockerImages"
```
jacek-izykowski: This directory is actually going to contain dockerfile (not docker image).
Better name could also mention that it is a "remote" location.

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.Tests.ps1
```diff
@@ -0,0 +1,19 @@
+Param (
```
jacek-izykowski: If DockerImages directory doesn't belong to tests, then why do these (IisTcpTestDockerImageBuild.Tests.ps1 and IisTcpTestDockerImageBuild.ps1) files are placed here (in Test/Utils)? We suggest moving powershell scripts to directory with dockerfile.

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.ps1
```diff
@@ -0,0 +1,18 @@
+. $PSScriptRoot\..\..\Common\Aliases.ps1
+$DockerfilePath = "$PSScriptRoot\..\..\DockerImages\iis-tcptest\Dockerfile"
+function Initialize-IisTcpTestDockerImage  {
```
jacek-izykowski: We suggest a blank line before 'function' block.

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.Tests.ps1
```diff
@@ -0,0 +1,19 @@
...
+    [Parameter(Mandatory=$true)] [string] $TestenvConfFile
+)
+
+. $PSScriptRoot\IisTcpTestDockerImageBuild.ps1
+
+$Sessions = New-RemoteSessions -VMs (Read-TestbedsConfig -Path $TestenvConfFile)
+$Session = $Sessions[0]
+Describe "Initialize-IisTcpTestDockerImage" {
```
jacek-izykowski: We suggest a blank line before 'describe' block.

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.Tests.ps1
katrybacka: What if we want to build this image for another purpose, for example on Linux? I think that Dockerfiles should be seperated from platform-specific scripts which use them. We use PS scripts only for tests but Dockerfiles could be used in other ways.

krdln: > What if we want to build this image for another purpose, for example on Linux?

I'd say that in that case, scripts for Linux can live alongside ps1 files. And the Dockerfiles are still separated – they live in their specific directories. Similar pattern exists in the `vtest` directory – this directory contains both `ps1` and `sh` scripts for testing, and actual tests are in `vtest/tests` directory. Perhaps this layout would make more sense?
```
CIScripts/
└── Docker
    ├── Dockerfiles
    │   └── iis-tcptest.Dockerfile
    └── Initialize-IisTcpTest.ps1
```

For me, the most important argument for moving the `ps1` file is to move related things together, so the reader of `Initialize-IIsTcpImage` doesn't have to move to totally different place of the repository to locate the Dockerfiles.

Anyway, I don't feels strongly about this directory structure, and if you think it's a good design, it could stay as it is. @jacek-izykowski thoughts?

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.Tests.ps1
```diff
@@ -0,0 +1,24 @@
...
+        { Initialize-IisTcpTestDockerImage -Session $Session } | Should Not Throw
+
+        Invoke-Command -Session $Session {
+            docker inspect iis-tcptest
+        } | Should Not BeNullOrEmpty
+    }
+
+    BeforeEach {
+        Invoke-Command -Session $Session {
+            docker image rm iis-tcptest -f
```
krdln: This line seems to fail if the image is not present. To silence the error you'd need to wrap it in a block with `SilentlyContinue` and add stderr redirection (note to self: we should allow invoking commands on remote session in `Invoke-NativeCommand`):
```
        Invoke-Command -Session $Session {
            Invoke-Command {
                $ErrorActionPreference = 'SilentlyContinue'
                docker image rm iis-tcptest -f 2>$null
            }
        }
```

CIScripts/Test/Utils/IisTcpTestDockerImageBuild.Tests.ps1
jacek-izykowski: Katarzyna, you may follow your path.

krdln:
I really like the parametrization of the `Initialize-DockerImage` function!

Minor nit (just for the future, don't bother changing now): If the function contains the word `Docker`, I think it doesn't make sens to name a parameter `-DockerImageName`. Just `-ImageName` would be perfectly suffiecient.

PR #53, opened 2018-03-22
define testEnvConfPath in 'provision' stage

sodar:


sodar:
recheck no bug

krdln:
Recheck no bug

PR #57, opened 2018-03-23
Disabling ipv6 at controller 

PawelMarchewkaCodiLime:


ansible/roles/yum-repos-prepare/tasks/main.yaml
```diff
@@ -1,5 +1,16 @@
 ---
 
+- name: 'copy apache config'
+  register: apache_config
```
PawelMarchewkaCodiLime: Correct this

ansible/roles/yum-repos-prepare/files/yum.conf
```diff
@@ -0,0 +1,27 @@
...
+debuglevel=2
+logfile=/var/log/yum.log
+exactarch=1
+obsoletes=1
+gpgcheck=1
+plugins=1
+installonly_limit=5
+bugtracker_url=http://bugs.centos.org/set_project.php?project_id=23&ref=http://bugs.centos.org/bug_report_page.php?category=yum
+distroverpkg=centos-release
+ip_resolve=IPv4
```
PawelMarchewkaCodiLime: I have added only this

PawelMarchewkaCodiLime:
It should help for http://10.84.12.26:8080/job/WinContrail/job/contrail-win-ci-gh/973/console

krdln:
MK + MJ: In general, we're :+1:. There's just one little corner case we're not sure can happen:

Currently, this `.conf` file has only one section: `[main]`. The line, if not present, will be added to the end of the file and will be properly inserted into the `[main]` section. However, if for whatever reasons, there will be another section `[whatever]` present in that file, the line on the end of file will be added to a wrong section.

So, if that scenario with multiple sections is possible, then perhaps we should use [`ini_file` module](https://web.archive.org/web/*/http://docs.ansible.com/ansible/latest/ini_file_module.html)? Otherwise, if the `[main]` is always the only section, it can stay as it is now.

Diabelko:
Does any of the tests use IPv6 at all? If not, I believe it should by done by disabling IPv6 in the kernel (preferably in the slave image).

PawelMarchewkaCodiLime:
@Diabelko I will try to disable IPv6 in the kernel, however I don't like idea of setting up this in the temple image. I think it would be better if it was expressed in our ansible playbooks.

PR #64, opened 2018-03-29
Implement PesterLogger.

m-kostrzewa:


CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
```diff
@@ -0,0 +1,137 @@
...
+
+function ItBlockOutsideOfDescribe() {
+    It "getting Pester scope works in It block outside of Describe" {
+        Get-CurrentPesterScope | Should -BeExactly @("PesterLogger", "It blocks in weird places",
+            "getting Pester scope works in It block outside of Describe")
+    }
+}
+
+Describe "PesterLogger" {
+    Context "when inside Context block" {
```
krdln: I'd put all the tests of `Get-CurrentPesterScope` in a separate `Context`, or perhaps even `Describe`. The tests should serve as examples, but now the reader is directly greeted with testing implementation details of `Get-CurrentPesterScope`

krdln:
I think a function such as `Copy-ToLogDir` would be a nice addition. It could be used to copy docker driver/agent logs to appropriate directory. Do you plan something like that for a followup PR?

krdln:
Could you explain how the `-OutDir` parameter would be set in our test scenarios? Where would that value come from?

m-kostrzewa:
@krdln 
Usage example:
```
*** PesterHelpers.Tests.ps1 ***
. $PSScriptRoot\..\..\Common\Init.ps1

. $PSScriptRoot\..\PesterLogger\PesterLogger.ps1
New-PesterLogger -OutDir "."

$here = Split-Path -Parent $MyInvocation.MyCommand.Path
$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.'
. "$here\$sut"

Describe "PesterHelpers" {
    Context "Consistently" {
        It "works on trivial cases" {
....
```
```
*** PesterHelpers.ps1 ***
function Eventually {
...
    Write-Log "test"
    Invoke-UntilSucceeds `
            -ScriptBlock $ScriptBlock `
            -AssumeTrue `
            -Interval $Interval `
            -Duration $Duration `
            -Name "Eventually"
}
```

CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
```diff
@@ -0,0 +1,58 @@
+﻿$here = Split-Path -Parent $MyInvocation.MyCommand.Path
+$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.'
+. "$here\$sut"
+
+Describe "PesterLogger" {
```
jacek-izykowski: Can we have a test case that actually invokes Add-ContentForce?
It may also be useful to verify that Push-Location and Pop-Location do not interfere with scope handling/file path.

CIScripts/Test/PesterLogger/PesterLogger.ps1
```diff
@@ -0,0 +1,35 @@
+﻿. $PSScriptRoot/Get-CurrentPesterScope.ps1
+
+function New-PesterLogger {
+    Param([Parameter(Mandatory = $true)] [string] $Outdir,
+          $WriterFunc = (Get-Item function:Add-ContentForce))
+
+    $DeducerFunc = Get-Item function:Get-CurrentPesterScope
```
jacek-izykowski: Comment explaining why do we need DeducerFunc might be useful here.

CIScripts/Test/PesterLogger/PesterLogger.ps1
```diff
@@ -0,0 +1,35 @@
...
+    }
+    Add-Content -Path $Path -Value $Value
+}
+
+function Register-NewWriteLogFunc {
+    Param($Func)
+    if (Get-Item function:Write-Log -ErrorAction SilentlyContinue) {
+        Remove-Item function:Write-Log
+    }
+    New-Item -Path function:\ -Name Global:Write-Log -Value $Func
```
jacek-izykowski: Shall we prevent New-Item output from appearing in stdout?

CIScripts/Test/PesterLogger/PesterLogger.ps1
```diff
@@ -0,0 +1,35 @@
...
+        & $WriterFunc -Path $Outpath -Value $Message
+    }.GetNewClosure()
+
+    Register-NewWriteLogFunc -Func $WriteLogFunc
+}
+
+function Add-ContentForce {
+    Param([string] $Path, [string] $Value)
+    if (-not (Test-Path $Path)) {
+        New-Item -Force -Path $Path -Type File
```
jacek-izykowski: Shall we prevent New-Item output from appearing in stdout?

CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
```diff
@@ -0,0 +1,58 @@
...
+
+    function MakeFakeWriter {
+        $Script:Content = ""
+        $Script:Path = ""
+        Get-Item function:InMemoryAddContent
+    }
+
+    Context "New-PesterLogger" {
+        It "registers a new global Write-Log function" {
+            New-PesterLogger -OutDir "some_dir"
```
krdln: I think that the `Set` or `Register` verb would be more appropriate here, as `New` usually the created resource, which is not a case here.

m-kostrzewa: Gonna go with Initialize

m-kostrzewa:
FYI I'm starting to use this logger in this PR: https://github.com/Juniper/contrail-windows-ci/pull/67

m-kostrzewa:
recheck no bug

PR #73, opened 2018-04-03
Implement Move-Logs to move logs from remote services.

m-kostrzewa:
Based on https://github.com/Juniper/contrail-windows-ci/pull/64

`Move-Logs` should be used in the highest-level `AfterEach` blocks, so that it get executed for every single test case.


m-kostrzewa:
recheck no bug

PR #74, opened 2018-04-04
Separate container creation in Single Host Utils tests

sodar:


CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1
```diff
@@ -130,9 +134,22 @@ Describe "Single compute node protocol tests with utils" {
...
     AfterEach {
-        Write-Host "Removing containers"
-        if (Get-Variable Container1ID -ErrorAction SilentlyContinue) {
+        function Test-IfVariableAndContainerExist {
+            Param ([Parameter(Mandatory=$true)] $ContainerID)
+
+            return $(
+                $(Get-Variable $ContainerID -ErrorAction SilentlyContinue) -and
+                $(docker ps -aq | Select-String $(Get-Variable $ContainerID))
+            )
```
krdln: All these `$()` subexpressions could be changed to just `()` parentheses. Also, the outer `()` is not necessary.


> ```powershell
> Select-String $(Get-Variable $ContainerID)
> ```

From what I've tested, this won't work, as `Get-Variable` returns an object. Try either

```powershell
Select-String $ContainerID
```

(`-and` is lazily evaluated, so missing variable on rhs won't cause strict mode to scream) or

```powershell
Select-String (Get-Variable ContainerID).Value
```

krdln:
Perhaps we can use `Invoke-NativeCommand -Session $Session` here? [See the tests for example usage](https://github.com/Juniper/contrail-windows-ci/blob/d3d495f18daf21954d2e7b64eaab1c679870bde5/CIScripts/Common/Invoke-NativeCommand.Tests.ps1#L32)

michal-clapinski:
recheck no bug

michal-clapinski:
recheck no bug

CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1
```diff
@@ -103,11 +104,17 @@ Describe "Single compute node protocol tests with utils" {
             -Name $NetworkName `
             -Subnet "$( $Subnet.IpPrefix )/$( $Subnet.IpPrefixLen )"
 
-        Write-Host "Creating containers"
-        $Container1ID, $Container2ID = Invoke-Command -Session $Session -ScriptBlock {
+        Write-Host "Creating container 1"
+        $Cmd1 = Invoke-NativeCommand -Session $Session -CaptureOutput {
```
PawelMarchewkaCodiLime: [Refinement] Do we need two variables Cmd1 and Cmd2? Maybe we could use only one variable like  $Command?

CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1
```diff
@@ -130,15 +137,26 @@ Describe "Single compute node protocol tests with utils" {
     }
 
     AfterEach {
-        Write-Host "Removing containers"
-        if (Get-Variable Container1ID -ErrorAction SilentlyContinue) {
-            Invoke-Command -Session $Session -ScriptBlock { docker rm -f $Using:Container1ID } | Out-Null
-            Invoke-Command -Session $Session -ScriptBlock { docker rm -f $Using:Container2ID } | Out-Null
+        Write-Host "Removing container 1"
+        if (Get-Variable "Container1ID" -ErrorAction SilentlyContinue) {
+            Invoke-NativeCommand -Session $Session -CaptureOutput {
```
PawelMarchewkaCodiLime: I wonder if we still need ```| Out-Null``` here, since when I try to do something like this:
```Invoke-NativeCommand -CaptureOutput {whoami.exe}``` I get some output in my console.

File: CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1 [hunk 4]
```diff
@@ -130,15 +137,26 @@ Describe "Single compute node protocol tests with utils" {
...
+                docker rm -f $Using:Container1ID
+            }
```
PawelMarchewkaCodiLime: We think that you should remove -CaptureOutput if you would like to print out result of `docker rm` or add `| Out-Null` after `Invoke-NativeCommand` to drop result of command.

michal-clapinski: `| Out-Null` is not needed

michal-clapinski:
recheck no bug

CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1
michal-clapinski: I don't really like reusing vars

CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1
krdln: `Invoke-NativeCommand` will return an object with `Output` field containing the output, which by default goes to the output stream. Do you want to explain why `Out-Null` is not needed here?

michal-clapinski: It does not go to the output stream by default. I don't know why and I don't care why, but it does not.

michal-clapinski: There a difference whether you run it in interactive mode or not.

michal-clapinski:
recheck no bug

PR #78, opened 2018-04-05
Move vars from gui to vault

katrybacka:


katrybacka:
recheck no bug

katrybacka:
recheck no bug

PR #85, opened 2018-04-09
Clean up and speed up Get-RemoteContainerNetAdapterInformation

krdln:


CIScripts/Test/Utils/CommonTestCode.ps1
```diff
@@ -63,27 +63,25 @@ function Get-RemoteContainerNetAdapterInformation {
...
-            MACAddressWindows = $MACAddressWindows;
-            IPAddress = $IPAddress;
-        }
+        $RemoteCommand = {
+            $Adapter = (Get-NetAdapter -Name 'vEthernet (Container NIC *)')[0]
+            @{
+                IfIndex = $Adapter.IfIndex;
+                IfName = $Adapter.IfName;
+                AdapterFullName = $Adapter.Name;
+                AdapterShortName = [regex]::new('vEthernet \((.*)\)').Replace($Adapter.Name, '$1')
```
michal-clapinski: Please move this after `Invoke-Command`

CIScripts/Test/Utils/CommonTestCode.ps1
```diff
@@ -63,27 +63,25 @@ function Get-RemoteContainerNetAdapterInformation {
...
+                MacAddressWindows = $Adapter.MacAddress.ToLower();
+                MacAddress = $Adapter.MacAddress.ToLower().Replace('-', ':')
```
michal-clapinski: Same as above

krdln: You mean to return only a partial struct and fill the rest of fields outside the `Invoke-Command`? I tried to do that before, but for some reason the Json returned by `ConvertFrom-Json` is immutable, and some [additional conversion would need to be performed](https://stackoverflow.com/a/40495378) or all the fields of struct would need to be set manually. So I thought it would be cleaner to just set it inside the remote command. Could you explain why do we want it to be after `Invoke-Command`?

michal-clapinski: For two reasons:
1. You copy-paste code, eg. `$Adapter.MacAddress.ToLower()`. If it was after `Invoke-Command` then you could use already existing fields.
2. It's weird. We should do as little as possible on remote machines and even less in containers on remote machines. This is post processing and I think it should be done where the rest of the code is executed.
You can use always some other serialization methods instead of JSON.

PR #67, opened 2018-03-30
Use Write-Log and RemoteLogCollector.

m-kostrzewa:
Puts detailed logs into detailed_logs directory (using `Write-Log` function and `RemoteLogCollector`).

Just look at it: http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/1364/detailed_logs/

http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/1364/detailed_logs/Single%20compute%20node%20protocol%20tests%20with%20utils.TCP%20connection%20works.log

PR #87, opened 2018-04-10
Implement RemoteLogCollector to move logs from remote services.

m-kostrzewa:
Reopening https://github.com/Juniper/contrail-windows-ci/pull/73 due to rebase difficulties

Move-Logs should be used in the highest-level AfterEach blocks, so that it get executed for every single test case. This is being done in https://github.com/Juniper/contrail-windows-ci/pull/67 that depends on this PR
(edit: I've halted working on that PR because it was getting harder and harder to rebase all the time.)


CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -0,0 +1,71 @@
+function New-TrackedLogSource {
+    Param([Parameter(Mandatory = $true)] [string] $Path,
+          [Parameter(Mandatory = $false)] [PSSessionT[]] $Sessions)
+
+    $Sources = @()
+
+    $Sessions | ForEach-Object {
```
sodar: We have thought of two alternatives for this loop.

One uses `foreach`: 

```powershell
If (-not $Sessions) {
  $Sessions = @($null)
}
foreach ($Session in $Sessions) {
  ...
}
```

Second one (it makes script analyzer suppression unnecessary):

```powershell
$Sources = $Sessions | ForEach-Object {
   ...

  $SingleSource
}
```

What do you think about them?

CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
```diff
@@ -0,0 +1,150 @@
...
+        "remote log text" | Should -BeIn $Content
+    }
+
+    It "cleans logs in source directory" {
+        $Source1 = New-TrackedLogSource -Sessions @($Sess1) -Path $DummyLog1
+        Initialize-PesterLogger -OutDir "TestDrive:\" -LogSources @($Source1)
+
+        Merge-TrackedLogs
+
+        Test-Path $DummyLog1 | Should -Be $false
```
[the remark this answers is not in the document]
m-kostrzewa: Trick here is that every test is invoked on local machine. $Sess1 can be $null or localhost. There are no tests on actual remote machines. For this reason this seems unnecessary.

CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
```diff
@@ -0,0 +1,150 @@
...
+        $ContentRaw | Should -BeLike "*$DummyLog1*remote log*"
+        $ContentRaw | Should -BeLike "*$DummyLog2*another file content*"
+    }
+
+    It "works with multiple sessions in single log source" {
+        $Source2 = New-TrackedLogSource -Sessions @($Sess1, $Sess2) -Path $DummyLog1
+        Initialize-PesterLogger -OutDir "TestDrive:\" -LogSources @($Source2)
+        Write-Log "first message"
+
+        Merge-TrackedLogs -DontCleanUp
```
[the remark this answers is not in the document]
m-kostrzewa: I've added a comment with explanation.

m-kostrzewa:
recheck no bug

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -0,0 +1,90 @@
...
+        Invoke-Command $Func -ArgumentList $Arguments
+    }
+}
+
+function Get-LogContent {
+    param([System.Collections.Hashtable] $LogSource)
+    $ContentGetterBody = {
+        Param([Parameter(Mandatory = $true)] [string] $From)
+        $Files = Get-ChildItem -Path $From -ErrorAction SilentlyContinue
+        $Logs = @{}
```
krdln: I don't see a reason for a hashmap here. Why not a list (eg. of tuples/hashmaps)?

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -0,0 +1,90 @@
...
+
+function New-LogSource {
+    Param([Parameter(Mandatory = $true)] [string] $Path,
+          [Parameter(Mandatory = $false)] [PSSessionT[]] $Sessions)
+
+    return $Sessions | ForEach-Object {
+        @{
+            Session = $_
+            Path = $Path
+        }
```
krdln: Perhaps make it a (data-only) class?

CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
```diff
@@ -0,0 +1,164 @@
+$here = Split-Path -Parent $MyInvocation.MyCommand.Path
+$sut = (Split-Path -Leaf $MyInvocation.MyCommand.Path) -replace '\.Tests\.', '.'
+. "$here\$sut"
+
+Describe "RemoteLogCollector" {
+    It "appends collected logs to correct output file" {
+        $Source1 = New-LogSource -Sessions @($Sess1) -Path $DummyLog1
```
krdln: Protip: You can pass a `T` where argument expects a `T[]` – ie. just `-Sessions $Sess1` – to avoid ascii-fair.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -1,71 +1,90 @@
...
-        $Closure = {
-            Invoke-Command $Func -ArgumentList $Arguments
-        }.GetNewClosure()
+        $Logs = Get-LogContent -LogSource $LogSource
+        $Logs.Keys | ForEach-Object {
+            $Filename = $_
+            $Content = $Logs[$Filename]
+            $SourceFilenamePrefix = "Contents of $($Filename):"
+
+            Write-Log ((@("-") * $SourceFilenamePrefix.Length) -join "")
```
sodar: `"-" * x`

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -1,71 +1,90 @@
...
+          [Parameter(Mandatory = $false)] [switch] $DontCleanUp)
 
-        [Diagnostics.CodeAnalysis.SuppressMessageAttribute(
-            "PSUseDeclaredVarsMoreThanAssignments", "Sources",
-            Justification="Seems like ForEach block messes up static analysis here - this var is" +
-                "already declared before the loop"
-        )]
-        $Sources += $SingleSource
-    }
+    $LogSources | ForEach-Object {
```
sodar: Since `$LogSources` will always have at least one element, maybe we should use `foreach` here, instead of `ForEach-Object`.

Please change this in other places as well.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
m-kostrzewa: Classes introduce problems with importing stuff, as I learned. I try to avoid them now.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
m-kostrzewa: I don't see any benefit of going with list + pairs here either and I don't think it's that important in this case. Hashmap feels more natural to me personally: each file maps to its content (hashmap) vs there are pairs of files and their contents. Also it guarantees uniqueness, although it's not really necessary in this case.

m-kostrzewa:
recheck no bug

PR #92, opened 2018-04-12
Add telegraf to builders and testers

katrybacka:


katrybacka:
recheck no bug

ansible/roles/telegraf/templates/telegraf.conf.j2
```diff
@@ -0,0 +1,262 @@
...
+  ## Logging configuration:
+  ## Run telegraf in debug mode
+  debug = false
+  ## Run telegraf in quiet mode
+  quiet = false
+  ## Specify the log file name. The empty string means to log to stdout.
+  logfile = "./telegraf.log"
+
+  ## Override default hostname, if empty use os.Hostname()
+  hostname = "host2"
```
krdln: This line should probably be commented out.

ansible/roles/telegraf/templates/telegraf.conf.j2
```diff
@@ -0,0 +1,262 @@
...
+# win_perf_counters over WMI is recommended)
+
+# # Read metrics about cpu usage
+# [[inputs.cpu]]
+#   ## Whether to report per-cpu stats or not
+#   percpu = true
+#   ## Whether to report total system cpu stats or not
+#   totalcpu = true
+#   ## Comment this line if you want the raw CPU time metrics
+#   fielddrop = ["time_*"]
```
krdln: I think that these `[[inputs.*]]` sections (without `win_perf_counters`) could be remove

PR #95, opened 2018-04-12
Implement CI selfcheck script.

m-kostrzewa:


m-kostrzewa:
recheck no bug

krdln:
I like it! There are a few things I'm not sure about though:
1. The tests are grouped into three categories: `CI_Unit`, `CI_Systest` and no-label. Are we sure we need all these three categories? If yes, I'd like to bikesched on naming scheme (the names are bikeschedable too)

   This PR | Proposal 1 | Proposal 2
   -- | -- | -- 
   `CI_Unit` | `CI` | no tag
   `CI_Systest` | `CI` + `Remote` | `Remote`
   no tag | `Remote` / no tag | `Real` / `Product` (or something like that)

   I feel the current proposal is kind of backwards (`CI_Unit` is quite a lot annotation for a simplest kind of test) and typo-prone
2. You've made the `TestenvConfFile` parameter not mandatory. I hope the error message (if you forget to pass it) is clear, ie. not somewhere deep in callstack. Have you checked that?
3. `-SkipUnit` & `-NoStaticAnalysis` — some inconsistency in naming here.

m-kostrzewa:
@krdln I also don't like the `CI_` prefix. However, I had to differentiate product tests until they are moved to other repo/directory.

Proposal `CI + Remote` can't work because `Invoke-Pester -Tags A, B` will run tests that have either A or B tag (OR, not AND).

For these reasons, I decided to go with `CI_` prefix for now. After test split I want to change it, however.

Invoke-Selfcheck.ps1
```diff
@@ -0,0 +1,52 @@
...
+# A new developer should be able to run `.\Invoke-Selfcheck.ps1` and it should pass 100% of the time,
+# without any special requirements, like libraries, testbed machines etc.
+# Special flags may be passed to invoke more complicated tests (that have requirements), but
+# the default should require nothing.
+
+
+$nl = [Environment]::NewLine
+
+function Get-Separator {
+    $nl + ((@("=") * 80) -join "") + $nl
```
sodar: It can be done more simply: `"=" * 80`.

Invoke-Selfcheck.ps1
```diff
@@ -0,0 +1,52 @@
...
+}
+
+function Write-VisibleMessage {
+    param([string] $Message)
+    Write-Host "$(Get-Separator)[Selfcheck] $Message$(Get-Separator)"
+}
+
+if ($SkipUnit) {
+    Write-VisibleMessage "-SkipUnit flag set, skipping unit tests"
```
sodar: Why did you even create a function like `Write-VisibleMessage`. What's wrong with normal `Write-Host`? Please don't write code for inserting 80 equals signs.

CIScripts/Test/Tests/Integration/DockerDriver.Tests.ps1
```diff
@@ -1,12 +1,9 @@
 Param (
-    [Parameter(Mandatory=$true)] [string] $TestenvConfFile
+    [Parameter(Mandatory=$false)] [string] $TestenvConfFile
```
sodar: Why did you make this parameter not mandatory? System tests require a testenv.

If config file has to be optional, then please check if `$TestenvConfFile` was passed in tests and fail them if it was not.

CIScripts/Test/Tests/Integration/DockerDriver.Tests.ps1
```diff
@@ -54,6 +51,15 @@ function Save-DockerDriverUnitTestReport {
 $modules = @("agent")
 
 Describe "Docker Driver" {
```
sodar: In tests for `Initialize-DockerImage` you have added ` -Tags CI_Systest`. Please add this tag to all system tests.

If it has to be this way, then why?

m-kostrzewa: CI selfcheck shouldn't include product systests. Initialize-DockerImage is in utils, so I figured it's part of CI selfcheck.

CIScripts/Test/Tests/Integration/DockerDriver.Tests.ps1
m-kostrzewa: Invoke-Pester runs the file first (asking for a Mandatory param) and THEN checks whether the tags are even there. If `Mandatory=$true` is set to true, then you get a lot of errors like "required parameter not specified", even for tests that you want to skip.

Invoke-Selfcheck.ps1
m-kostrzewa: I added separators because I wanted it to standout more from Pester reports.

I'm going to remove `Get-Separator` func because of this cool syntax `"=" * 80` but I think `Write-VisibleMessage` should stay. In other case, every `Write-Host` would have to add separators and `[Selfcheck]` prefix which looks ugly.

CIScripts/Test/Tests/Integration/DockerDriver.Tests.ps1
m-kostrzewa: I don't have a better idea to solve this except to mark everything as not mandatory, and check in runtime

krdln:
@m-kostrzewa 
> Proposal CI + Remote can't work because Invoke-Pester -Tags A, B will run tests that have either A or B tag (they use OR, not an AND

You can just run
```powershell
Invoke-Pester -Tags CI
```
if testenv config is present and
```powershell
Invoke-Pester -Tags CI -ExcludeTag Remote
```
if not.

m-kostrzewa:
@krdln Oh, solution with ExcludeTag makes sense

m-kostrzewa:
To reviewers: I merged in development branch, please recheck your comments (github has hidden them)

CIScripts/Test/Tests/Integration/DockerDriver.Tests.ps1
m-kostrzewa: (went with @krdln 's suggestion here)

m-kostrzewa:
recheck no bug

m-kostrzewa:
recheck no bug

PR #96, opened 2018-04-13
Retry starting docker driver also on invalid IP

krdln:
This PR enables retrying on "Waiting for IP on interface time out"<sup>†</sup> error. Also, we do now print the cause of retry when retrying.

The reason for this error is probably some HNS flakiness. It should be fixed directly in Docker Driver, along with the "HNS unspecified" error. But since we're retrying in CI anyway, I see no reason why we shouldn't retry also on this error.

The actual change is in second commit – https://github.com/Juniper/contrail-windows-ci/commit/35d2e0d4f498229667973b50d21083b1860cef8d. The rest are minor bugfixes and cleanup.

 † The _time out_ is a little bit misleading here. Since we no longer have a DHCP and have manual configuration, the correct IP will never appear, no matter how long we wait. When the error occurs, the assigned IP is autoconfiguration one, instead of the same address as on Ethernet1 – eg. 172.16.0.2.

### DIFF
--- a/CIScripts/Common/Invoke-UntilSucceeds.Tests.ps1
+++ b/CIScripts/Common/Invoke-UntilSucceeds.Tests.ps1
@@ -156,7 +156,23 @@ Describe "Invoke-UntilSucceeds" {
         } -Duration 10 -Interval 5
 
         $Script:Counter | Should Be 2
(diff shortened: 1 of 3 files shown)


krdln:
Recheck no bug

PR #100, opened 2018-04-16
Move old tests directory.

m-kostrzewa:
Reasoning:

i've been trying to move some files around and these just get in the way.


m-kostrzewa:
recheck no bug

PR #102, opened 2018-04-17
Run RemoteLogCollector tests in selfcheck script.

m-kostrzewa:


CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
```diff
@@ -146,19 +173,45 @@ Describe "RemoteLogCollector" {
...
+    }
+
+    AfterEach {
+        Remove-Item "TestDrive:/*" 
+        if (Get-Item function:Write-Log -ErrorAction SilentlyContinue) {
+            Remove-Item function:Write-Log
         }
     }
+
+    Test-MultipleSourcesAndSessions -DescribeName "RemoteLogCollector - with actual Testbeds"
```
PawelMarchewkaCodiLime: We wonder if there could be a way to get DescribeName. Is it possible to use `Get-CurrentPesterScope` here?

PR #107, opened 2018-04-18
Enabling multi-node Agent tests

jacek-izykowski:


jacek-izykowski:
recheck no bug

jacek-izykowski:
recheck no bug

PR #109, opened 2018-04-18
Start telegraf service

jablonskim:


katrybacka:
recheck no bug

ansible/roles/telegraf/tasks/main.yml
```diff
@@ -14,5 +14,10 @@
 - name: Add telegraf to PATH
   win_path:
     elements:
-    - 'C:\Program Files\telegraf'
+    - '{{ telegraf_path }}'
     state: present
+
```
katrybacka: What do you think about moving the starting of telegraf to deploy task, so the template wouldn't send statistics to graphite?

PR #110, opened 2018-04-19
Build, deploy testenv and perform sanity tests concurrently

PawelMarchewkaCodiLime:


Jenkinsfile
```diff
@@ -78,130 +77,123 @@ pipeline {
...
+                        }
+                    }
                 }
-            }
-        }
 
-        stage('Cleanup-Provision-Deploy-Test') {
-            agent none
-            when { environment name: "DONT_CREATE_TESTBEDS", value: null }
+                stage('Cleanup-Provision') {
```
PawelMarchewkaCodiLime: I think better name would be just "Provision". It is going to look better in Jenkins. I will a comment that cleanup is done before provisioning.

Jenkinsfile
```diff
@@ -78,130 +77,124 @@ pipeline {
...
-
-                        ansibleExtraVars = vmwareConfig + testEnvConfig
-
-                        // 'Cleanup' stage
-                        node(label: 'ansible') {
                             deleteDir()
                             unstash 'Ansible'
 
                             dir('ansible') {
+                                // Cleanup testenv before making a new one
```
m-kostrzewa: This makes me wonder...

Originally, we wanted to destroy testenv in `post` stage. We moved it earlier because of issues with holding a lock for so long.

In this proposed PR, lock is pretty much global. Does this mean we can move testenv destruction back into `post` stage? Do we want to? Depending on how long testenv provisioning takes compared to build - it might be beneficial.

Jenkinsfile
```diff
@@ -1,14 +1,13 @@
...
 
-def ansibleExtraVars
-
 pipeline {
     agent none
 
     options {
         timeout time: 5, unit: 'HOURS'
         timestamps()
+        lock label: 'testenv_pool', quantity: 1
```
m-kostrzewa: Also wondering, whether there's a way to monitor how many locks are used. And if we want to monitor this.

Jenkinsfile
PawelMarchewkaCodiLime: I think it should be doable. However it would include some refactor since we would need `ansibleExtraVars`  also in `post` stage. Moreover I think it might be a bit risky if the job would assume that testenv is already cleared. I don't want to do that in this PR. However yes it should be beneficial.

Jenkinsfile
PawelMarchewkaCodiLime: If there was an issue with testenv_pool we would see job queue growing.

PawelMarchewkaCodiLime:
a) Yes, locking is working. My jobs with this change have been run simultaneously with other jobs and there was no issue with sharing the same testenv_pool.
b) Yes lock is released please take a look at http://10.84.12.26:8080/job/WinContrail/job/WinCI-pm-test/49/console
c) Here we have two concurrent jobs WinCI-pm-test as you can see they have adjured different testenv_pools:
- http://10.84.12.26:8080/job/WinContrail/job/WinCI-pm-test/54/execution/node/135/log/
- http://10.84.12.26:8080/job/WinContrail/job/WinCI-pm-test/53/execution/node/135/log/

Yes `failFast` policy will require additional logic, since vmware task cancellation is not instant.

PR #117, opened 2018-04-24
Linked clones: Create snapshot

jablonskim:


ansible/roles/vmware-template/tasks/shutdown.yml
```diff
@@ -0,0 +1,27 @@
+---
+
+- name: Power Off Template-{{ vm_role }}
+  vmware_guest:
+    hostname: "{{ vcenter_hostname }}"
+    username: "{{ vcenter_user }}"
+    password: "{{ vcenter_password }}"
+    validate_certs: "{{ validate_certs }}"
+    name: "{{ new_vm.instance.hw_name }}"
```
sodar: Name can be removed, because UUID has already been passed.

PR #120, opened 2018-04-25
Use Remove-AllContainers in tests

krdln:


CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -470,13 +470,15 @@ function Remove-Container {
 }
 
 function Remove-AllContainers {
-    Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session)
-
-    Invoke-Command -Session $Session -ScriptBlock {
-        $Containers = docker ps -q
```
jacek-izykowski: We suggest using "-aq" to list also containers that are not running.

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -470,13 +470,15 @@ function Remove-Container {
...
-        if($Containers) {
-            docker rm -f $Containers | Out-Null
+    Param ([Parameter(Mandatory = $true)] [PSSessionT[]] $Session)
```
jacek-izykowski: Suggested refinement: In all other places, where we accept one or more sessions, we call this parameter `Sessions` (e.g. Deployment.ps1, RemoteLogCollector.ps1).

krdln: As described in commit message, I went with powershell naming convention (see the `-Session` param [here](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/remove-pssession?view=powershell-6)), but I can change it.

PR #122, opened 2018-04-25
Don't use deprecated contrail-generateds repo

PawelMarchewkaCodiLime:
Use only contrail-api-client


PawelMarchewkaCodiLime:
recheck no bug

PawelMarchewkaCodiLime:
recheck no bug

PR #126, opened 2018-04-26
Capture containers' logs

krdln:


CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -41,23 +65,58 @@ function Get-LogContent {
...
+    }
+}
+
+function Get-LogContent {
+    param([LogSource] $LogSource)
+
+    if ($LogSource.Source['Path']) {
+        Get-FileLogContent -Session $LogSource.Session -Path $LogSource.Source.Path
+    } elseif ($LogSource.Source['Container']) {
+        Get-ContainerLogContent -Session $LogSource.Session -Container $LogSource.Source.Container
```
m-kostrzewa: Maybe something akin to a class would be better here? LogSource class with two subtypes: File and Container, with methods Clear and Get-Content.

1. Class inheritance would work given all those classes (LogSource, File and Container) are in the same file.
2. Otherwise, New-*LogSource functions could return structs with fields pointing to relevant functions (like in one of early prototypes of RemoteLogCollector)

I wouldn't do any of the above if it's too problematic though.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -1,9 +1,10 @@
+. $PSScriptRoot/../../Common/Aliases.ps1
 . $PSScriptRoot/../../Common/Invoke-NativeCommand.ps1
 
 . $PSScriptRoot/PesterLogger.ps1
 
 class LogSource {
-    [PSSessionT] $Session
+    [System.Management.Automation.Runspaces.PSSession] $Session
```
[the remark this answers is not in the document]
krdln: harold

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
krdln: I've tried to simulate tagged union with match in powershell and perhaps it may look weird. I'd argue that it's readable enough.

Solution 2. would have low readability (hard to track where's the code that actually runs), so I'd prefer not to do it.

Solution 1. seems nice, but do we want everybody to learn yet another feature of powershell just to do a simple switch? I'll try to implement it using classes but not sure if it's worth the time.

m-kostrzewa: > tagged union 
Very procedural, not OOP. However, without classes we probably are better off sticking to it. Notice how adding a new log source method requires changing `function Get-LogContent`. It should've been handled by polymorphysm. But classes in PS don't work too well so.. meh.

> learn yet another feature of powershell just to do a simple switch

but this is not a valid reason imo.

> Solution 2. would have low readability (hard to track where's the code that actually runs), so I'd prefer not to do it.

This is a tradeoff between maintainability and explicitness. Switch-case is more explicit but harder to maintain in the long run, because adding (extending) the system requires changes in unrelated (client) parts. 

I believe by removing the "closures" that were in the original (rejected) implementaiton of RemoteLogCollector, and now by using a switch-case, we're introducing a tech debt. Imagine this switch-case having 10 fields, for different log sources... A nightmare.

The only reason it's OK in this case is because powershell polymorphysm sucks.

CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
```diff
@@ -211,10 +212,32 @@ Describe "RemoteLogCollector - with actual Testbeds" -Tags CI, Systest {
...
+        BeforeEach {
+            Initialize-PesterLogger -OutDir "TestDrive:\"
+        }
+
+        It "captures logs of container" {
+            New-Container -Session $Sess1 -Name foo -Network nat
+
+            Merge-Logs (New-ContainerLogSource -Session $Sess1 -Container foo)
+            $ContentRaw = Get-Content -Raw "TestDrive:\*.Docker logs.captures logs of container.log"
+            $ContentRaw | Should -BeLike "*Microsoft Windows*"
```
m-kostrzewa: Could you provide an example log with two types of log sources?

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -1,11 +1,94 @@
...
+    }
+}
+
+function New-FileLogSource {
     Param([Parameter(Mandatory = $true)] [string] $Path,
           [Parameter(Mandatory = $false)] [PSSessionT[]] $Sessions)
 
     return $Sessions | ForEach-Object {
-        @{
+        [FileLogSource] @{
```
jacek-izykowski: Possible refinement: what do you think about writing a constructor?

CIScripts/Test/Tests/Protocol/SingleHostWithUtils.Tests.ps1
```diff
@@ -137,7 +137,14 @@ Describe "Single compute node protocol tests with utils" {
                 Remove-Variable "ContrailNetwork"
             }
         } finally {
-            Merge-Logs -LogSources (New-LogSource -Path (Get-ComputeLogsPath) -Sessions $Session)
+            [LogSource[]] $LogSources = New-FileLogSource -Path (Get-ComputeLogsPath) -Sessions $Session
+            foreach ($ContainerVar in "Container1ID", "Container2ID") {
```
jacek-izykowski: This procedure seems to be more complicated than necessary. Can we handle container names/ids in the same manner as in TunnellingWithAgent.Tests.ps1? In those tests container names are fixed.

CIScripts/Test/Tests/Protocol/TunnellingWithAgent.Tests.ps1
```diff
@@ -482,7 +482,10 @@ Describe "Tunnelling with Agent tests" {
             Clear-TestConfiguration -Session $Sessions[0] -SystemConfig $SystemConfig
             Clear-TestConfiguration -Session $Sessions[1] -SystemConfig $SystemConfig
         } finally {
-            Merge-Logs -LogSources (New-LogSource -Path (Get-ComputeLogsPath) -Sessions $Sessions)
+            [LogSource[]] $LogSources = New-FileLogSource -Path (Get-ComputeLogsPath) -Sessions $Sessions
+            $LogSources += New-ContainerLogSource -Session $Sessions[0] -Container $Container1ID
```
jacek-izykowski: `Session` or `Sessions`? `Container` or `ContainerNames`? Please check parameters' names for New-ContainerLogSource. We know that Powershell can handle parameter name abbreviations, but please use full names as they are declared.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -1,11 +1,94 @@
...
+        $Command = Invoke-NativeCommand -Session $this.Session -CaptureOutput -AllowNonZero {
+            docker logs $Using:this.Container
+        }
+        return @{
+            "$( $this.Container ) container logs" = $Command.Output
+        }
+    }
+}
+
+function New-ContainerLogSource {
```
jacek-izykowski: It may actually return multiple elements. Function name may be misleading. The same remark applies to New-FileLogSource.

krdln: This is a standard practice in powershell, for example, `New-Item` usually returns just one element, but
```powershell
New-Item -Type file -Path "foo", "bar", "baz"
```
creates multiple files.

There's even a lint in powershell analyzer that forbids plural nouns (but it's disabled in our CI)

CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
```diff
@@ -211,10 +212,32 @@ Describe "RemoteLogCollector - with actual Testbeds" -Tags CI, Systest {
...
+        }
+
+        It "handles nonexisting container" {
+            Merge-Logs (New-ContainerLogSource -Session $Sess1 -Container bar)
```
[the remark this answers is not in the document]
krdln: The problem with:
```powershell
{ Something } | Should -Not -Throw
```
is that in case of an exception being thrown Pester doesn't show the exception and its location at all.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
krdln: I'm against writing constructors for plain-data classes.

CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
krdln: Do you mean a test?

krdln:
I've just noticed that `Merge-Logs` should be called before `Remove-AllContainers` as the logs are empty now, so I'll need to reorder some stuff.

CIScripts/Test/PesterLogger/RemoteLogCollector.Tests.ps1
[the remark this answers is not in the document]
m-kostrzewa: Well that's dumb. In this case 
```
{ Something } # | Should -Not -Throw
# Assert is commented out because Pester doesn't show the exception and its location at all
```
would be better

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
m-kostrzewa: Except it has two methods, it's not plain data at all? :confused:

m-kostrzewa: (but whatever, it's possible refinement so I approve)

krdln: You're right. I think I've forgotten that I had refactored it to *have* methods.

Anyway, if the constructor does nothing more than setting fields to arguments, I see no point in creating it. Especially since this class is an implementation detail (though I don't know if you can mark something as private in PS), and the `New-FileLogSource` is the user-facing constructor.

m-kostrzewa: We were wondering for calling constructor if e.g. only one remote session is used, and calling `New-*LogSource`*s* when multiple are. But since you pointed out that convention is not to have plurals for cmdlet names, then I think this is a moot point and I agree.

PR #130, opened 2018-04-27
Use Write-Log instead of Write-Host in all tests.

krdln:
(most importantly, in the Agent UDP test)
Also, clean up RemoveAllContainers.Tests so it no longer depends on docker driver.


CIScripts/Test/Utils/CommonTestCode.ps1
```diff
@@ -1,4 +1,5 @@
 . $PSScriptRoot\..\..\Common\Aliases.ps1
+. $PSScriptRoot\..\Test\PesterLogger\PesterLogger.ps1
```
krdln: This is added as a preparation for #127

krdln:
> Write-Log function could be moved from PesterLogger.ps1 to CIScripts/Test/Utils/Write-Log.ps1 or something. And it's the one that should be included everywhere where it's used.

I'll do it in the follow-up PR

PR #137, opened 2018-05-08
Removed symlinking boost and MSBUILD variable

PawelMarchewkaCodiLime:
During a build we use boost which is already installed on builder
MSBUILD variable is already set on builders

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -90,7 +90,6 @@ pipeline {
                         CERT_PASSWORD_FILE_PATH = "C:/BUILD_DEPENDENCIES/third_party_cache/common/certs/certp.txt"
                         COMPONENTS_TO_BUILD = "DockerDriver,Extension,Agent"
 
-                        MSBUILD = "C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe"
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -90,7 +90,6 @@ pipeline {
                         CERT_PASSWORD_FILE_PATH = "C:/BUILD_DEPENDENCIES/third_party_cache/common/certs/certp.txt"
                         COMPONENTS_TO_BUILD = "DockerDriver,Extension,Agent"
 
-                        MSBUILD = "C:/Program Files (x86)/MSBuild/14.0/Bin/MSBuild.exe"
```
michal-clapinski: Not sure about that. Maybe we should leave it here and remove it from ansible?

PawelMarchewkaCodiLime: I believe it is better to have this variable set in ansible, where `VisualCPP Build Tools` are installed. 

Imagine someone would like to move MSBuild.exe to another location, consequently we would have to update MSBUILD. It is better to do that in one step when ansible playbook is run. Synchronizing new value of MSBUILD in `jenkinsfile` may be a bit more complicated.

michal-clapinski: no i przekonał!

PR #141, opened 2018-05-10
Migrate VM-Template to NFS-Datastore

jablonskim:


jablonskim:
recheck no bug

jablonskim:
recheck no bug

ansible/vmware-create-template.yml
```diff
@@ -27,6 +27,14 @@
         stage: snapshot
       when: vm_role == "testbed"
 
+    - name: Migrate VM
+      include_role:
+        name: vmware-template
+      vars:
+        stage: migrate
+        vcenter_datastore_name: "NFS-Datastore"
```
sodar: We should extract this into a playbook parameter.

ansible/vmware-create-template.yml
```diff
@@ -32,7 +32,7 @@
         name: vmware-template
       vars:
         stage: migrate
-        vcenter_datastore_name: "NFS-Datastore"
+        vcenter_datastore_name: "{{ vcenter_datastore_name }}"
```
sodar: You can remove this line, since `vcenter_datastore_name` is provided as extra variable.

PR #142, opened 2018-05-11
Add linked clone

katrybacka:
TODO:
- [ ] Use datastore cluster (from storage_pod object) as requested.
- [ ] Use wintb templates created by @jablonskim 
- [ ] Create ctrl template for linked clones and use in deploy-testenv playbooks (maybe another pr)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -142,8 +142,9 @@ pipeline {
 
                     environment {
                         TESTBED = credentials('win-testbed')
-                        TESTBED_TEMPLATE = "Template-testbed-201804050628"
-                        CONTROLLER_TEMPLATE = "Template-CentOS-7.4-Thin"
+                        // TODO: change after creation of proper templates
+                        TESTBED_TEMPLATE = "tmpl-wintb-master"
+                        CONTROLLER_TEMPLATE = "tmpl-ctrl-master"
                         TESTENV_MGMT_NETWORK = "VLAN_501_Management"
                         TESTENV_FOLDER = "WINCI/testenvs"
                         VCENTER_DATASTORE_CLUSTER = "WinCI-Datastores-SSD"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -143,8 +143,8 @@ pipeline {
                     environment {
                         TESTBED = credentials('win-testbed')
                         // TODO: change after creation of proper templates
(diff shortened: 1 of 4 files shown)


ansible/roles/vmware-vm/tasks/deploy.yml
```diff
@@ -7,7 +7,7 @@
       - "--user '{{ vcenter_username }}'"
       - "--password '{{ vcenter_password }}'"
       - "--datacenter '{{ vcenter_datacenter }}'"
-      - "--datastore-cluster '{{ vcenter_datastore_cluster }}'"
+      - "--cluster '{{ vcenter_cluster }}'"
```
sodar: I think it would be better to choose datastores from given datastore cluster, instead of the whole cluster.

Jenkinsfile
```diff
@@ -116,8 +116,9 @@ pipeline {
 
                     environment {
                         TESTBED = credentials('win-testbed')
-                        TESTBED_TEMPLATE = "Template-testbed-201804050628"
-                        CONTROLLER_TEMPLATE = "Template-CentOS-7.4-Thin"
+                        // TODO: change after creation of proper templates
+                        TESTBED_TEMPLATE = "tmpl-wintb-master"
+                        CONTROLLER_TEMPLATE = "tmpl-ctrl-master"
```
sodar: Since we already have some testbed templates created by @jablonskim , maybe we can plug them here already? What do you think about this?

Also we would need to create separate CentOS template.

ansible/files/vmware_api.py
```diff
@@ -69,6 +70,21 @@ def get_vm_folder(self, folder_path):
...
+        if len(datastores) == 0:
+            return None, None
+        random.shuffle(datastores)
+        chosen_datastore = datastores[0]
+        chosen_host = chosen_datastore.host[0].key
+        return chosen_host, chosen_datastore
+
+
+    def _get_datastores_from_cluster(self):
+        datastores = [d for d in self.datacenter.datastore if 'ssd' in d.name]
```
sodar: Please see a comment regarding choosing cluster. Storage pod object (http://pubs.vmware.com/vsphere-6-5/index.jsp#com.vmware.wssdk.apiref.doc/vim.StoragePod.html) has references to its datastores.

PR #150, opened 2018-05-21
[WIP] Enable CI selfcheck and logs collection in CI.

m-kostrzewa:


Jenkinsfile
```diff
@@ -218,17 +242,24 @@ pipeline {
...
+                            -OutputDir TestReports/WindowsCompute'''
 
-                        dir("processed_reports") {
-                            stash name: 'processedTestReport', allowEmpty: true
-                        }
+                        powershell script: '''./CIScripts/GenerateTestReport.ps1 `
+                            -XmlsDir testReportsRaw/CISelfcheck `
+                            -OutputDir TestReports/CISelfcheck'''
+
+                        // Using robocopy to workaround 260 chars path length limitation.
```
krdln: Could you calrify where is this limitation? This comment doesn't seem to explain that.

Jenkinsfile
```diff
@@ -218,17 +242,24 @@ pipeline {
...
+                        // TODO: Similar method may be used when CISelfcheck generates detailed logs.
+                        robocopy("${pwd()}/testReportsRaw/WindowsCompute/detailed/", "${pwd()}/TestReports/WindowsCompute/detailedLogs", "*.log")
```
krdln: The return status of this [`robocopy`] function is ignored.
Also, this command produces lot of output, perhaps we should hide it?

Jenkinsfile
m-kostrzewa: https://github.com/Juniper/contrail-windows-ci/pull/166

PR #153, opened 2018-05-23
Fix driver junit report discovery. Refactors.

aethyrxeo:
Collect docker driver unit tests xml junit reports from testbed machines. Move them to a temporary reports location for now (to avoid any problems with html report generation). Some slight refactors along the way.


michal-clapinski:
recheck no bug

michal-clapinski:
recheck no bug

PR #166, opened 2018-05-28
Fix robocopy output handling and verbosity.

m-kostrzewa:
Addressing @krdln 's comments under https://github.com/Juniper/contrail-windows-ci/pull/150/files


krdln:
The robocopy could be replaced by `mv`. Should be fixed in followup PR.

m-kostrzewa:
recheck no bug

PR #171, opened 2018-05-29
Disallow '/' (and other characters) in test names and fix them

krdln:


CIScripts/Test/PesterLogger/PesterLogger.ps1
```diff
@@ -24,11 +24,10 @@ function Initialize-PesterLogger {
         Param([Parameter(Mandatory = $true)] [object] $Message)
         $Scope = & $DeducerFunc
         $Filename = ($Scope -join ".") + ".log"
-        $Outpath = Join-Path $Script:ConstOutdir $Filename
-        if ($Outpath -Like "*:*:*") {
-            # We check for more than one ":" character. (There is always one, after drive letter)
-            throw [UnsupportedPesterTestNameException] "Invalid test name; it cannot contain some special characters, like ':'"
+        if (($Filename.IndexOfAny([System.IO.Path]::GetInvalidFileNameChars())) -ne -1) {
```
m-kostrzewa: Hmm, this is getting more complicated. Maybe we should move this check along with unit tests to a separate module?

krdln: I think it'd be overkill, this is a standard method of validating filenames in Powershell. Verbose? Yes. Complicated? Meh. Perhaps just extracting to a function would be ok?

krdln:
recheck no bug

PR #173, opened 2018-06-04
Fail when multiple IP addresses are returned for single Net Adapter

jablonskim:


CIScripts/Test/Utils/CommonTestCode.ps1
```diff
@@ -69,11 +77,15 @@ function Get-RemoteContainerNetAdapterInformation {
...
             $Adapter | Select-Object $Fields | ConvertTo-Json
-        }.toString()
+        }.ToString()
 
         docker exec $Using:ContainerID powershell $RemoteCommand
     } | ConvertFrom-Json
 
+    if ($Adapter.IPAddress -is [array]) {
+        throw "Multiple IP addresses returned from container for single NetAdapter."
+    }
```
krdln: Do you think printing these addresses here would be valueable?

krdln: [refinement] Do you think printing these addresses here would be valuable?

jablonskim:
recheck no bug

PR #178, opened 2018-06-07
Add single retry to New-Container

jacek-izykowski:


CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+            -Session $Session `
+            -NetworkName "NetworkOfFriends" `
+            -Name "jolly-lumberjack" | Should -Throw
+    }
+
+    It "Reports container id when container creation succeeds in second attempt after failing because of known issue" {
+        Invoke-Command -Session $Session {
+            $DockerThatSucceedsInSecondAttempt = @'
+            if ($args[0] -eq "run") {{
+                $Hope = "HopeForSuccess"
```
jablonskim: a more descriptive var name please

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+                if (Test-Path $Hope) {{
+                    Write-Output "{0}"
+                    Remove-Item $Hope
+                    exit 0
+                }} else {{
+                    Set-Content -Path $Hope -Value "New hope"
```
jablonskim: empty file would be simpler

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+                    Write-Output "{0}"
+                    Write-Error "It's Docker here: CreateContainer: failure in a Windows system call. Try again. Good luck!"
```
jablonskim: message should be as simple as possible, eg. `CreateContainer: failure in a Windows system call` and nothing else

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -469,7 +470,22 @@ function New-Container {
     if ($Name) { $Arguments += "--name", $Name }
     $Arguments += "--network", $NetworkName, $Image
 
-    $ContainerID = Invoke-Command -Session $Session { docker @Using:Arguments }
+    $Result = Invoke-NativeCommand -Session $Session -CaptureOutput -AllowNonZero { docker @Using:Arguments }
```
jablonskim: Please use dependency injection to provide a string to docker executable. By default: `docker.exe` and in tests `TestDrive:\docker.ps1` for example

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+                    Set-Content -Path $Hope -Value "There's actually no hope."
+                    Write-Output "{0}"
+                    Write-Error "It's Docker here: Error code #190583. Shall you try again?"
+                    exit 1
+                }}
+            }} else {{
+                Write-Output "{0}"
+                exit 0
+            }}
+'@ -f $Using:ContainerID,$HopeStorage
```
jablonskim: Decide on one mechanism for supplying the path - either hardcoded or passed

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+            -Session $Session `
+            -NetworkName "SoftwareDefinedNetwork" `
+            -Name "jolly-lumberjack" | Should -Not -Throw
+        $NewContainerID | Should -Be $ContainerID
+    }
+
+    It "Throws exception when container creation fails in first attempt and reports unknown issue" {
+        Invoke-Command -Session $Session {
+            $HopeStorage = "HopeForLuckyTry"
+            Remove-Item $HopeStorage -ErrorAction Ignore
```
jablonskim: You're removing this file before the test here, but not in the other test. Why?

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+            if ($args[0] -eq "run") {{
+                $Hope = "{1}"
+                if (Test-Path $Hope) {{
+                    Write-Output "{0}"
+                    Remove-Item $Hope
+                    exit 0
+                }} else {{
+                    Set-Content -Path $Hope -Value "There's actually no hope."
+                    Write-Output "{0}"
+                    Write-Error "It's Docker here: Error code #190583. Shall you try again?"
```
jablonskim: I'd prefer `unknown error` or something as simple as that

CIScripts/Test/NewContainer.Tests.ps1
jablonskim: Maybe use [this](https://github.com/pester/Pester/wiki/TestDrive) for all files created in tests

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+            Remove-Item $HopeStorage -ErrorAction Ignore
+            $DockerThatSucceedsInSecondAttempt = @'
+            if ($args[0] -eq "run") {{
+                $Hope = "{1}"
+                if (Test-Path $Hope) {{
+                    Write-Output "{0}"
+                    Remove-Item $Hope
+                    exit 0
+                }} else {{
+                    Set-Content -Path $Hope -Value "There's actually no hope."
```
jablonskim: again, empty file preferred

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,127 @@
...
+        )]
+        $Session = $Sessions[0]
+        [Diagnostics.CodeAnalysis.SuppressMessageAttribute(
+            "PSUseDeclaredVarsMoreThanAssignments", "",
+            Justification="Lifetime and visibility of variables is a matter beyond capabilities of code checker."
+        )]
+        $ContainerID = "47f6baf1e42fa83b5ddb6a8dca9103178129ce454689f47ee59140dafc2c9a7c"
+        Invoke-Command -Session $Session {
+            $OldPath = $Env:Path
+            $Env:Path = ".;$OldPath"
```
jablonskim: Remove this after implementing dependency injection

CIScripts/Test/NewContainer.Tests.ps1
jacek-izykowski: Then we are going to have a special case when pattern is the same as a string that we search in.

CIScripts/Test/TestConfigurationUtils.ps1
jacek-izykowski: I already use dependency injection - it's just on a bit different level. It's also more clean, because tested code doesn't need to have special features to make tests possible.

Adding/removing a single file on testbed also doesn't seem to matter. Our tests are already doing much more serious operations - installing msis, starting containers, etc. And TestDrive is another thing to care about when one tests the code without Pester.

CIScripts/Test/NewContainer.Tests.ps1
jacek-izykowski: It's hardcoded when there is only one piece of code that is using a path - as simple as possible.
It's passed as a variable when it's used in multiple places.

CIScripts/Test/NewContainer.Tests.ps1
jacek-izykowski: This is dependency injection.

CIScripts/Test/NewContainer.Tests.ps1
```diff
@@ -0,0 +1,130 @@
...
+    [Parameter(Mandatory=$false)] [string] $LogDir = "pesterLogs",
+    [Parameter(ValueFromRemainingArguments=$true)] $AdditionalParams
+)
+
+. $PSScriptRoot\TestConfigurationUtils.ps1
+. $PSScriptRoot\..\Testenv\Testenv.ps1
+. $PSScriptRoot\..\Testenv\Testbed.ps1
+. $PSScriptRoot\PesterLogger\PesterLogger.ps1
+
+Describe "New-Container" {
```
m-kostrzewa: Could you tag this as CI and Systest?
```
Describe "New-Container" -Tags CI, Systest {
```

m-kostrzewa: ( so that Invoke-CISelfcheck executes this)

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -473,7 +474,24 @@ function New-Container {
...
+    $ContainerID = $Result.Output.Split([Environment]::NewLine) | Select-Object -First 1
+    $OutputMessages = $Result.Output
+
+    # Workaround for occasional failures of container creation in Docker for Windows.
```
m-kostrzewa: Just two comments:
* can we add this to HNS bestiary?
* maybe we should move comment + workaround code to separate function(s)?

CIScripts/Test/NewContainer.Tests.ps1
jacek-izykowski: In separate PR. We want it to pass CI without these tags.

CIScripts/Test/TestConfigurationUtils.ps1
jacek-izykowski: And TestDrive is available in session in which Pester is invoked. And this code is writing files in remote session.

PR #180, opened 2018-06-08
Add unit tests for Wait-RemoteInterfaceIP

aethyrxeo:
So we know that "waiting for ip on interface..." bug isn't residing at CI side


aethyrxeo:
Opinion on variables names/test file name/tests description?

CIScripts/Test/TestConfigurationUtils.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        It "AddressFamily isn't matching" {
+            $MockedGetNetIPAddress.AddressFamily = "IPv6"
+            $TestResult = $MockedGetNetIPAddress | Select-ValidNetIPInterface
+            $TestResult | Should BeNullOrEmpty
+        }
+
+        It "SuffixOrigin isn't matching" {
+            $MockedGetNetIPAddress.SuffixOrigin = "WellKnown", "Link", "Random"
```
krdln: Is this `$MockedGetNetIPAddress` a copy of `$MockedGetNetIPAddress` defined in `BeforeEach` or is it actually a global variable?

Anyway, it would be better to avoid sharing global state between tests and construct a fresh object each time.

CIScripts/Test/TestConfigurationUtils.Tests.ps1
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        It "Both AddressFamily and SuffixOrigin are empty strings" {
+            $MockedGetNetIPAddress.AddressFamily = ""
+            $MockedGetNetIPAddress.SuffixOrigin = ""
+
+            $TestResult = $MockedGetNetIPAddress | Select-ValidNetIPInterface
+            $TestResult | Should BeNullOrEmpty
+        }
+
```
krdln: It would be also nice to have a test in which there are multiple objects passed to `Select-ValidNetIPInterface` (eg. the first one valid but second one not, and vice versa). I've recently seen a weird behaviour when function that uses `ValueFromPipeline` was looking only at the last value from list and I needed to use [`Process {}` block](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_functions_advanced_methods?view=powershell-6#process).

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -304,8 +314,7 @@ function Wait-RemoteInterfaceIP {
         Invoke-Command -Session $Session {
             Get-NetAdapter -Name $Using:AdapterName `
                 | Get-NetIPAddress -ErrorAction SilentlyContinue `
-                | Where-Object AddressFamily -eq IPv4 `
-                | Where-Object { ($_.SuffixOrigin -eq "Dhcp") -or ($_.SuffixOrigin -eq "Manual") }
+                | Select-ValidNetIPInterface
```
krdln: You can try the following approach:
```powershell
$Body = (Get-Item function:Select-ValidNetIPInterface).Definition
Invoke-Command -Session $Session {
    function Select-ValidNetIPInterface {
        & [Sriptblock]::Create($Using:Body)
    }
    ...
}
```

But perhaps that's an overkill and another place for errors to occur.

Perhaps we could make a wrapper:
```powershell
Invoke-CommandWithFunctions -Session $Session `
    -Functions "Select-ValidNetIPInterface" { ... }
```
that would handle copying functions to remote session for us and be properly tested. But perhaps that's even bigger overkill.

CIScripts/Test/TestConfigurationUtils.Tests.ps1
krdln: Sorry for the comment above, I was somehow thinking you're doing `BeforeAll`, not `BeforeEach`.

CIScripts/Test/TestConfigurationUtils.Tests.ps1
```diff
@@ -3,46 +3,82 @@
...
+                AddressFamily = "IPv4";
+                SuffixOrigin = @("Dhcp", "Manual")
+            }
+            $MockedGetNetIPAddress = $InvalidGetNetIPAddress, $ValidGetNetIPAddress
+
+            $TestResult = $MockedGetNetIPAddress | Select-ValidNetIPInterface
+            $TestResult | Should Be $ValidGetNetIPAddress
+        }
+
+        It "Pass invalid then valid object into pipeline" {
```
krdln: These could be grouped into single `It` to avoid duplication, but meh.

CIScripts/Test/TestConfigurationUtils.ps1
```diff
@@ -297,22 +297,27 @@ function Remove-AllUnusedDockerNetworks {
...
 function Wait-RemoteInterfaceIP {
     Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session,
            [Parameter(Mandatory = $true)] [String] $AdapterName)
+    $InjectedFunction = [PSCustomObject]@{ Name = 'Select-ValidNetIPInterface'; Body = ${Function:Select-ValidNetIPInterface} }
 
-    Invoke-UntilSucceeds -Name "Waiting for IP on interface $AdapterName" -Duration 60 {
+    Invoke-UntilSucceeds -Name "Waiting for IP on interface $AdapterName" -Duration 15 {
         Invoke-Command -Session $Session {
-            $ScriptBlock = [ScriptBlock]::Create($Using:SelectValidNetIPInterfaceBody)
+            $Using:InjectedFunction | ForEach-Object { Invoke-Expression "function $($_.Name) { $($_.Body) }" }
```
krdln: [possible future refinement] Wrap this in some kind of `Invoke-CommandWithCaptureFunctions` helper with proper tests etc.

[nit] When using `$( expr )` expression substitution in strings, it's nice to put spaces inside `()`, ie. `$( $_.Name )`. (Not sure where I've read that style rule, but I believe it's not something that I've just invented)

aethyrxeo:
Tests passed so this can be reviewed

CIScripts/Test/TestConfigurationUtils.Tests.ps1
```diff
@@ -0,0 +1,82 @@
...
+            $MockedGetNetIPAddress = @{
+                AddressFamily = "";
+                SuffixOrigin = ""
+            }
+
+            $TestResult = $MockedGetNetIPAddress | Select-ValidNetIPInterface
+            $TestResult | Should BeNullOrEmpty
+        }
+
+        It "Pass valid/invalid object combinations into pipeline" {
```
m-kostrzewa: I wonder if we can run all these tests using this method: https://github.com/Juniper/contrail-windows-ci/pull/58#discussion_r177790308 (Parametrized test cases)

It would remove some boilerplate

CIScripts/Test/TestConfigurationUtils.Tests.ps1
```diff
@@ -0,0 +1,82 @@
...
+            $ValidGetNetIPAddress = @{
+                AddressFamily = "IPv4";
+                SuffixOrigin = @("Dhcp", "Manual")
+            }
+
+            $MockedGetNetIPAddress = @($InvalidGetNetIPAddress, $ValidGetNetIPAddress)
+            $TestResult = $MockedGetNetIPAddress | Select-ValidNetIPInterface
+            $TestResult | Should Be $ValidGetNetIPAddress
+
+            $MockedGetNetIPAddress = $InvalidGetNetIPAddress, $ValidGetNetIPAddress
```
jacek-izykowski: It's exactly the same case as tested above.

CIScripts/Test/TestConfigurationUtils.Tests.ps1
```diff
@@ -0,0 +1,82 @@
+. $PSScriptRoot\TestConfigurationUtils.ps1
```
jacek-izykowski: What do you think about enabling strict mode? It can be done by invoking Init.ps1 (see other tests).

krdln:
@m-kostrzewa 
> I wonder if we can run all these tests using this method: #58 (comment) (Parametrized test cases)

[personal opinion (ignore if you want)]
I feel like the tests after your suggestion feel overengineered and less readable, especially the conditional logic inside the testcase: `if ($( $TestCase.Valid ))`. I'm inspired to add a warning in shelly about usage of `if` and `foreach` inside `It` 😄

Perhaps we could group only the invalid cases, leaving valid ones in separate tests? (so we can avoid conditional). Perhaps, we can ditch the `foreach`, ie.?

```powershell
@{ AddrFamily = Something; ... } | Select-ValidIpAddresses | Should -BeNullOrEmpty
@{ AddrFamily = SomethingElse; ... } | Select-ValidIpAddresses | Should -BeNullOrEmpty
```

I would suggest we could use the [`-TestCases` mechanism builtin into pester](http://mikefrobbins.com/2016/12/09/loop-through-a-collection-of-items-with-the-pester-testcases-parameter-instead-of-using-a-foreach-loop/), but I think some part of our logging/reporting pipeline doesn't support it.

@aethyrxeo 
[nit] _I think_ that in most cases you can replace `$( foo )` by just `foo` and `@( bar )` with just `(bar)` (the `@{}` needs to stay though).

aethyrxeo:
@jacek-izykowski @krdln I think it is better now.

PR #181, opened 2018-06-08
Add timestamps to pester logs and change them to .txt

krdln:


CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -18,26 +32,36 @@ class LogSource {
...
+                    Err = "<FILE NOT FOUND>"
+                }
             } else {
                 foreach ($File in $Files) {
                     $Content = Get-Content -Raw $File
-                    $Logs[$File.FullName] = if ($Content) {
-                        $Content
-                    } else {
-                        "<FILE WAS EMPTY>"
+                    $Logs += @{
```
jacek-izykowski: Why specific type is not available here? Why do we need casting from lines 57-64? Maybe a comment in the code might be beneficial for future users.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
```diff
@@ -125,12 +161,19 @@ function Merge-Logs {
         Write-Log ("=" * 100)
         Write-Log $ComputerNamePrefix
 
-        $Logs = $LogSource.GetContent()
-        foreach ($Log in $Logs.GetEnumerator()) {
-            $SourceFilenamePrefix = "Contents of $($Log.Key):"
-            Write-Log ("-" * 100)
-            Write-Log $SourceFilenamePrefix
-            Write-Log $Log.Value
+        foreach ($Log in $LogSources.GetContent()) {
```
jacek-izykowski: How does it work? You are iterating over LogSources in a loop that iterates over LogSources... A typo?

CIScripts/Test/PesterLogger/PesterLogger.ps1
```diff
@@ -21,25 +21,57 @@ function Initialize-PesterLogger {
     $ConstOutdir = Resolve-Path $Outdir
 
     $WriteLogFunc = {
-        Param([Parameter(Mandatory = $true)] [object] $Message)
+        Param(
+            [Parameter(Mandatory = $true)] [object] $Message,
+            [parameter(ValueFromRemainingArguments=$true)] $WriterArgs,
+            [Parameter(Mandatory=$false)] [string] $Tag = "tester",
```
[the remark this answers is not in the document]
krdln: I don't think this name would actually lead to confusion, especially when testers' hostnames are `winci-tester-x`, not just `tester`. I think of _tester_ as a just machine performing tests -- be it ci tester, laptop or something else.

If you insist of changing this default, what do you think about:
* Defaulting to hostname
* Not printing a label at all for default messages
* Changing the default to `pester`
* Changing the default to `root` (I can see this will be misleading and provoke a questions "what is a root?")

Changing the default to `default` is a bad idea imho as this name is totally meaningless for somebody unaware of how our logging is implemented.

> MKos: What do you think about adding specialized Host-LogSource?

I think this would make sense when two-stage `Collect` + `Merge-Logs` will be implemented, which we've agreed to be out of scope of this PR.

CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
```diff
@@ -28,35 +28,49 @@ Describe "PesterLogger" -Tags CI, Unit {
...
+            Write-Log $true
+            Get-Content "TestDrive:\PesterLogger.Write-Log.can write non-strings.txt" |
+                ConvertTo-LogWithoutTimestamps | Should -Be "tester | True"
+        }
+
+        It "can write arrays" {
+            Initialize-PesterLogger -OutDir "TestDrive:\"
+            Write-Log ("foo", "bar")
+            Get-Content "TestDrive:\PesterLogger.Write-Log.can write arrays.txt" |
+                ConvertTo-LogWithoutTimestamps | Should -Be ("tester | foo", "tester | bar")
```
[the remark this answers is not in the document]
krdln: Good idea! I was thinking about that but thought it would be an overkill, but after all it sounds reasonable.

CIScripts/Test/PesterLogger/RemoteLogCollector.ps1
krdln: This scriptblock is sometimes invoked on remote machine. You're right, I'll add a comment.

CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
```diff
@@ -68,13 +82,58 @@ Describe "PesterLogger" -Tags CI, Unit {
...
+                Write-Log "foo"
+                Get-Content "TestDrive:\PesterLogger.Write-Log.Timestamps.are present in logs.txt" |
+                    Should -Be "2018-06-08 14:46:44.042000 | test-runner | foo"
+            }
+
+            It "can be disabled" {
+                Initialize-PesterLogger -OutDir "TestDrive:\"
+                Write-Log -NoTimestamps "foo"
+                Get-Content "TestDrive:\PesterLogger.Write-Log.Timestamps.can be disabled.txt" |
+                    Should -Be "                           | test-runner | foo"
```
m-kostrzewa: We should probably use `ConvertTo-LogItem` here and in `It "are present in logs"` as well

CIScripts/Test/PesterLogger/PesterLogger.ps1
```diff
@@ -21,25 +21,57 @@ function Initialize-PesterLogger {
...
-    Add-Content -Path $Path -Value $Value | Out-Null
+
+    $TimestampFormatString = 'yyyy-MM-dd HH:mm:ss.ffffff'
+
+    $Prefix = if ($UseTimestamps) {
+        Get-Date -Format $TimestampFormatString
+    } else {
+        " " * $TimestampFormatString.Length
+    }
+    $Prefix += " | " + $Tag + " | "
```
m-kostrzewa: I wonder if we could/would want to extract lines 55-62 to a separate function

CIScripts/Test/PesterLogger/PesterLogger.ps1
```diff
@@ -59,3 +91,21 @@ function Write-Log {
...
+class LogItem {
+    [String] $Timestamp
+    [String] $Tag
+    [String] $Message
+}
+
+function ConvertTo-LogItem {
+    Param([Parameter(ValueFromPipeline, Mandatory=$true)] $Line)
+    Process {
+        $Timestamp, $Tag, $Message = $Line.Split("|", 3)
```
m-kostrzewa: Maybe we should add comments with example log lines (with and without a timestamp)?

CIScripts/Test/PesterLogger/PesterLogger.ps1
m-kostrzewa: (64-72 maybe too?)

m-kostrzewa:
[Uber-nitpick] Also just realized, that 
```
Context "Timestamps" 
```
is plural, but 
```
It "are present in logs"
```
is singular.

We should use singular `Context` or reword the `It`.

Also,  `It "can be disabled"` - refers to `Timestamps` - another mismatch.

CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
```diff
@@ -68,13 +82,58 @@ Describe "PesterLogger" -Tags CI, Unit {
...
+
+        Context "Tags" {
+            It "default to test-runner" {
+                Initialize-PesterLogger -OutDir "TestDrive:\"
+                Write-Log "foo"
+                Get-Content "TestDrive:\PesterLogger.Write-Log.Tags.default to test-runner.txt" |
+                    ConvertTo-LogItem | Foreach-Object Tag | Should -Be "test-runner"
+            }
+
+            It "can be set" {
```
m-kostrzewa: The [Uber-nitpick] applies to this `Context` block as well

CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
krdln: I think it's good to have a test that checks the exact format and I didn't want to create a separate one just for that.

CIScripts/Test/PesterLogger/PesterLogger.ps1
krdln: Calling external functions in the closure is quite annoying, as you have to explicitly capture it. Do you think it's worth doing?

CIScripts/Test/PesterLogger/PesterLogger.ps1
krdln: "The tests are examples"

krdln:
@m-kostrzewa 

You should treat `It` as a keyword, not exactly as specific pronoun, the test output reads very well:
```
      Context Timestamps
        [+] are present in logs 278ms
        [+] can be disabled 32ms
```

Singular `Timestamp` sound really weird as a context.

m-kostrzewa:
> You should treat It as a keyword

no, we shouldn't. It should read like English in code directly. Pester is constructed as a BDD framework, where test code should be as close to natural language as possible.

Ginkgo does reporting similarly:
```
VRouter
  errors if extension is not Running
  C:/go_workspace/src/github.com/Juniper/contrail-windows-docker-driver/core/vrouter/vrouter_test.go:45
```

Ginkgo has `It` and an alias `Specify`.  https://onsi.github.io/ginkgo/#individual-specs-it

I believe, that Pester should also have an alias for `It`. However, in many cases, like in the linked example:

```
Describe("The foobar service", func() {
  Context("when calling Foo()", func() {
    Context("when no ID is provided", func() {
      Specify("an ErrNoID error is returned", func() {
      })
    })
  })
})
```

.. it can also be worded using an `It`: `It("returns an error ID"...`

I mean, you could say that lacking of `Specify` alias is a flaw in Pester. But I think I disagree with your comment. `It`, `BeforeEach`, etc. are specifically called like that to be as close to natural language as possible.

CIScripts/Test/PesterLogger/PesterLogger.ps1
m-kostrzewa: Hmm are you sure that it needs to be done in this case? If it is required, then maybe adding a comment before each block would suffice

CIScripts/Test/PesterLogger/PesterLogger.Tests.ps1
m-kostrzewa: I disagree, just like you shouldn't test if your UI shows a red button or a green button. Formatting is meant to be changed.

Or at least have at most literally one test case for exact formatting method. But I expect it to change a lot and the test is fragile.

CIScripts/Test/PesterLogger/PesterLogger.ps1
m-kostrzewa: We shouldn't have an example of specific formatting, as mentioned in my earlier comment. Such test would be fragile and pointless.

If we had a complex regex here, and had 10 test cases to see if it matches correctly, I would still recommend having a "happy path example" just next to the regex. `This regex matches strings like asdfasdf`.

`ConvertTo-LogItem` exists only because we use actually files on TestDrive, instead of having a nice interface for the output method (we don't have something like the Emitter in Shelly, that can output to a vector or to stdout). `ConvertTo-LogItem` is pretty much a workaround for this fact. But it's a good workaround for now. 

However, trying to write tests that are coupled to specific formatting even more would move as away from the 'perfect' scenario of having replaceable output and formatting methods. We already have to rework `ConvertTo-LogItem` function every time we add a new tagging method (e.g. showing `timestamp | hostname | logname | msg`. Or `| priority |`. etc.

krdln:
@m-kostrzewa 
When I said “should be treated as a keyword” I meant “a keyword meaning 3rd person pronoun, regardless of the plularity” not just a “test goes here” keyword.

(and as an aside – actually I like the lack of `Specify` – it forces you to think of better names for tests :), or perhaps even better tests)

m-kostrzewa:
@krdln I mean, it's a nitpick, you can apply it if you want ;)

CIScripts/Test/PesterLogger/PesterLogger.ps1
krdln: Hah, I've forgotten I've already moved this logic out of the closure 😆

PR #183, opened 2018-06-11
Enable strict mode everywhere

krdln:
To enable strict mode, we've usually been using `Init.ps1` (which enables strict mode and sets error action). This PR adds this import in all root scripts and tests.

The exception is the `StaticAnalysis/Invoke-StaticAnalysisTools.ps1`—I've decided to set the strict mode explicitely, to keep it independent of `CIScripts`.

I thought that we were already using strict mode everywhere, but adding a warning for missing strict mode in [**shelly**](https://github.com/krdln/shelly) had proved me wrong.


krdln:
recheck no bug

PR #186, opened 2018-06-13
Move Builders and Testers to separate directories

jablonskim:


krdln:
Can you show an example of deployed tester / builder?

jablonskim:
For now: `ci-builder-9`

PR #188, opened 2018-06-14
Change name of winci-* VMs to fit in 15 char limit

jablonskim:


ansible/roles/vmware-template/tasks/deploy.yml
```diff
@@ -31,7 +31,7 @@
     hardware: "{{ vm_hardware }}"
     state: poweredon
     customization:
-      hostname: "winci-{{ vm_role }}-{{ get_vm_id.stdout.strip() }}"
+      hostname: "ci-{{ vm_role }}-{{ get_vm_id.stdout.strip() }}"
```
krdln: Can we do someting to avoid repetition here?

m-kostrzewa:
Just a suggestion: maybe write a check that the name is too long that will fail the playbook?

PR #205, opened 2018-06-26
Grep recheck

krdln:


krdln:
@sodar @m-kostrzewa I've added you to review, but I'm using ansible in a really hacky way, so treat the playbook as WIP for now, as I'm going to refactor it a little.

krdln:
@sodar I think it's reviewable now.

krdln:
recheck no bug

ansible/roles/post-recheck-comment/tasks/main.yaml
```diff
@@ -0,0 +1,64 @@
...
+  register: comments
+  raw: gerrit query change:{{ change_id }} --comments --format JSON
+  retries: 6
+  delay: 10
+  until: >
+    comments.rc == 0 and
+    (comments.stdout_lines[0] | from_json).comments
+    | selectattr('reviewer.username', 'equalto', ci_user)
+    | map(attribute='message')
+    | list
```
sodar: We think that we can improve readability of this conditional (and any other expression which includes comments) by employing a custom filter. Let's say this filter would be named `extract_messages`. `extract_message` would do the same as the following expression:

```
(comments.stdout_lines[0] | from_json).comments
| selectattr('reviewer.username', 'equalto', ci_user)
| map(attribute='message')
```

Play would then look like this:

```
 name: Wait for Verified-1 comment from {{ ci_user }}
 register: comments
 raw: gerrit query change:{{ change_id }} --comments --format JSON
 retries: 6
 delay: 10
 until: >
   comments.rc == 0 and
   comments | extract_messages | last | regex_search(failure_pattern)
```

Here you can find an example of a custom filter which we tested: https://gist.github.com/sodar/a530ad10cc81cbe47a494f398b0bfd31

What do you think about this approach?

ansible/roles/post-recheck-comment/tasks/main.yaml
```diff
@@ -0,0 +1,64 @@
...
+  set_fact:
+    n_failures: >
+      {{
+      (comments.stdout_lines[0] | from_json).comments
+      | selectattr('reviewer.username', 'equalto', ci_user)
+      | map(attribute='message')
+      | map('regex_search', failure_pattern)
+      | select
+      | list
+      | count
```
sodar: Since `n_failures` is used as an int later, couldn't we use `int` filter here?

Then `int` filter from next tasks could be removed.

sodar: This applies to `max_rechecks` as well.

ansible/roles/post-recheck-comment/tasks/main.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+- name: Show failure count
+  debug:
+    var: n_failures
+
+- name: Prepare the recheck message
+  when: n_failures | int < max_rechecks | int + 1
+  set_fact:
+    comment_message: >
+      We've detected that Windows CI failed due to flakiness, so we're going to recheck windows
```
sodar: We were thinking `n_failures` logic. Right now it counts how many `Verified: -1` comments were posted, but it takes into account each CI failure - flakiness and compilation errors for example.

Maybe we should calculate `n_failures` as an amount of comments containing `recheck windows`. What do you think?

ansible/roles/post-recheck-comment/tasks/main.yaml
```diff
@@ -0,0 +1,64 @@
...
+      | list
+      | count
+      }}
+
+- name: Show failure count
+  debug:
+    var: n_failures
+
+- name: Prepare the recheck message
+  when: n_failures | int < max_rechecks | int + 1
```
sodar: We could remove `+1` by simply doing the following:

```
n_failures <= max_rechecks
```

sodar: Also next condition based on `n_failures` and `max_rechecks` can be changed like this as well.

flakes/grep.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -e
+set -u
+set -o pipefail
+
```
sodar: We could add `set -x` to each bash script to print each executed command.

ansible/inventory.sample
```diff
@@ -15,12 +15,17 @@ drive
 ansible-worker
 jenkins-server
 
+[gerrit:children]
+zuul
```
sodar: `zuul` shouldn't be a part of `gerrit` group, because if `zuul` host would be in inventory then `post-recheck-comment` playbook would be run on gerrit server and zuul server as well.

We propose to define two groups:
- `gerrit` - only varialbes
- `gerrit-servers` - for gerrit hosts

Then we could inherit variables like this:

```
[gerrit:children]
gerrit-server
zuul
```

ansible/post-recheck-comment.yml
```diff
@@ -0,0 +1,68 @@
+---
+
+- hosts: localhost
```
sodar: We have an idea to reduce complexity of this playbook. Idea requires a few steps:

1. Define gerrit server in inventory directly
    - we could create `inventory.production` and `inventory.development`
    - production variant would hold `review.opencontrail.org`
    - development variant would hold `review2.opencontrail.org`
2. We could extract keyfile creating into a separate script/playbook
    - we were thinking about emploing `yq` tool and use it like this

        ```bash
        ansible-vault view group_vars/gerrit/vault | yq -r ".gerrit_keys.${GERRIT_HOST}" > "${KEY_FILE}"
        ```

    - then we could run a playbook `post-recheck-comment.yml` (we're only not sure about `extra-args` part)

        ```
        ansible-playbook -i inventory.production \
            --private-key="${KEY_FILE}" \
            --ssh-extra-args="-o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no" \
            post-recheck-comment.yml
        ```

    - this way `post-recheck-comment.yml` playbook would be responsible only for:
        - prompting for change id and patchset
        - running `post-recheck-comment` role

We checked `yq` tool availability and it kinda sucks because:

- it depends on `jq` tool, installable with apt
- `yq` can then be installed with `pip`

We have tested the keyfile extraction command from point 2. and it worked for following versions:

- `sudo apt install jq=1.5+dfsg-1`
- `pip install yq==2.6.0`

We can discuss IRL, but what do you think?

ansible/roles/post-recheck-comment/tasks/main.yaml
krdln: Funny fact – no. I've tested it and unfortunately somehow it gets stringified somewhere and doesn't work.

ansible/roles/post-recheck-comment/tasks/main.yaml
krdln: Good idea! I thought about using a custom filter but it seems I didn't google hard enough. The only small disadvantage is requiring user to jump between files and that it's not really obvious where `extract_messages` comes from (imagine reading this code and furiously trying to google `ansible extract_messages`). But with a comment, I think it would be ok.

flakes/grep.sh
krdln: I'm not a fan of `set -x` in scripts that has it's own output (I know it's only stderr), as it's noisy and might be confusing. Perhaps just let's jus `set -x` in `./run-tests.sh`?

ansible/inventory.sample
krdln: Agreed. Although using `gerrit` only for variables seems weird. Perhaps we can use the following hierarchy?

* `gerrit_variables`
    * `zuul`
    * `gerrit` (or `gerrit-server`)

ansible/post-recheck-comment.yml
krdln: Not sure about it. This indeed reduces scope of this playbook, but simultaneously adds some assumptions, moves some of the complexity to a separate script and makes it harder to run this playbook by hand.

> we could create `inventory.production` and `inventory.development`

This seems like a good idea. Is it possible to have the inventory sample, with group definitions, as we now have and create two files that "include" that inventory and just define gerrit servers?

> it depends on jq tool, installable with apt

Sidenote: Funny fact, I've used the `jq` for filtering jsons in the firsts iteration, but then rewritten using filters, because I didn't want to introduce a new dependency.

ansible/roles/post-recheck-comment/tasks/main.yaml
krdln: > flakiness and _compilation errors_ for example.

I'm not sure I understand, if the compilation error had occurred, then the author must have submitted a new patchset to fix it. And we're looking at the `-1` comments only for a specific patchset, so the older failures doesn't count at all.

flakes/grep.sh
krdln: I've tried doing `set -x` in run-tests and it's not really readable, so think I won't do it for now.

ansible/roles/post-recheck-comment/tasks/main.yaml
sodar: Yes, jumping between the files could cause problems, but in the end this conditional will be more understandable.

As to the naming, maybe we should namespace it? E.g. `cw_extract_messages` - contrail windows extract messages. Still, someone might google it 😄. Your call.

Also, IIRC `filter_plugins` directory can be either in root directory or in the role directory. Since it is used only here, maybe it should be in role directory.

ansible/roles/post-recheck-comment/tasks/main.yaml
sodar: Ok, makes sense. We misunderstood the part about patchsets.

ansible/inventory.sample
sodar: It might be okay. Other proposal is `gerrit-clients`. This way name of the group represents hosts' purpose.

krdln: Since now we have gerrit in the inventory, we can remove the `gerrit-variables` and make `zuul` pull all the needed variables from `gerrit`'s vars. But it's out of scope if this PR, I guess.

PR #210, opened 2018-06-29
zuul: setup-zuul-server playbook

sodar:
- Moved necessary variables to production inventory
- Playbook for setting up/updating zuul server


sodar:
recheck no bug

PR #213, opened 2018-07-02
zuul: move gerrit_known_hosts variables to gerrit-variables group

sodar:


sodar:
recheck no bug

sodar:
recheck no bug

PR #216, opened 2018-07-04
Fix testbed template creation

jacek-izykowski:


vars/prepareAndRunAnsiblePlaybook.groovy
```diff
@@ -20,7 +20,7 @@ def call(Map params) {
         agent { label 'ansible' }
         steps {
           dir('ansible') {
-            sh 'cp inventory.sample inventory'
+            sh 'cp inventory.groups inventory'
```
sodar: I just had an idea. I think that we could omit this copy, and just use `inventory.groups` directly in `ansiblePlaybook` in the step below.

jacek-izykowski:
recheck no bug

PR #217, opened 2018-07-04
Copy SCons logs to logserver

jablonskim:


jablonskim:
Currently there is only `kernel-tests` target in scons. In the future there will be Agent tests too

PR #223, opened 2018-07-09
flakes: TLS c-a-d network timeout

sodar:


sodar:
Closing due to #224 fixing the problem

PR #225, opened 2018-07-12
Rename ScontsTestsLogs to vRouterUnitTestReport

m-kostrzewa:
I feel like the old name is not descriptive enough for folks outside of Windows team.


m-kostrzewa:
recheck windows

m-kostrzewa:
recheck no bug

m-kostrzewa:
@jablonskim vRouter actually is often refered to by Juniper when talking about vRouter + Agent combo. What do you think then?

jablonskim:
Ok, then it's OK for me

m-kostrzewa:
@jablonskim k

PR #231, opened 2018-07-24
flakes: recheck on 'no route to host' error

sodar:
Example of a failed build: http://10.84.12.26:8080/job/WinContrail/job/winci-server2016-prod/6370/console

To fix this flakiness (rather rare flakiness) a change in contrail-ansible-deployer is needed. Namely, support for configuring custom docker registry for containers other than kolla/contrail is required. `redis` is an example of such container.


flakes/patterns.txt
```diff
@@ -5,3 +5,4 @@ CITimeoutException: Waiting for IP on interface Ethernet1 failed.
 SIGNTASK : SignTool error : No certificates were found that met all the given criteria\.
 \[[[:digit:]]+-controller\]: FAILED.*Failed to connect to yum.dockerproject.org at port 443
 ERROR:zuul\.Repo:Unable to initialize repo for https://review\.opencontrail\.org/Juniper/[a-z-]+\s*$
+Get https://registry-1\.docker\.io/v2/:.*getsockopt: no route to host
```
krdln: Maybe we should change `registry-1` to `registry-[[:digit:]]+`? From what I understand, there are multiple registry servers and you can be redirected to different one. (perhaps I'm wrong though)

sodar: I would stay with `registry-1`, because I did not find any information about registries other than `registry-1`. Also DNS properly resolves only for `registry-1.docker.io` from what I checked and `registry-1` itself resolves to multiple IP addresses.

sodar: But I'll add the flake description test file.

PR #234, opened 2018-07-25
logs: move SconsTestLogs to unittest-logs; index.html in pretty_test_report

sodar:


sodar:
Note to self:

- might be useful to extract testenv provisioning logs to a separate file
- pipeline stage logs can be extracted but length is limited by a configurable property in Jenkins plugin
    - log extracting example: https://gist.github.com/sodar/ba1cd102ad79fceddbc2bf835198d686
- property in question is `com.cloudbees.workflow.rest.external.FlowNodeLogExt.maxReturnChars`
    - configurable with `init.groovy` files
    - example: https://github.com/rancher/pipeline-jenkins-server/blob/0e4e1044a3e389784d68a1d9c66848b7a5f905c2/init.groovy

sodar:
recheck no bug

sodar:
recheck no bug

jablonskim:
recheck no bug

PR #235, opened 2018-07-26
Extract Test directory.

m-kostrzewa:
I finally did it


m-kostrzewa:
recheck no bug

Test/Rdir/raw_NUnit/report.xml
```diff
@@ -0,0 +1,257 @@
+﻿<?xml version="1.0" encoding="utf-8" standalone="no"?>
```
pmarcinkiewicz1: Uncle Bob would be very disappointed... Why do we need this file?

PR #237, opened 2018-07-27
Update Controller to ocata-master-206

jablonskim:


ansible/roles/contrail-ansible-deployer/tasks/main.yaml
```diff
@@ -48,6 +48,33 @@
     msg: configure_instances.yml playbook has failed
   when: configure_instances_command is failed
 
+- name: Provision instances with install_openstack.yml playbook
```
sodar: Most of this task file is repetition. We could extract calling playbook, registering output to a separate file. Maybe this would work:

```yaml
- include_tasks: 'run_deployer_playbook.yml'
  vars:
    playbook: '{{ item.playbook }}'
    extra_variables: '{{ item.extra_variables | default("") }}'
  with_items:
    - { playbook: 'playbooks/configure_instances.yml' }
    - { playbook: 'playbooks/install_openstack.yml' }
    - { playbook: 'playbooks/install_contrail.yml', extra_variables: '-e orchestrator=openstack' }
```

What do you think about this?

PR #243, opened 2018-07-30
Recheck broken powershell session error

krdln:


m-kostrzewa:
@krdln Dariusz is swamped so I reviewed it instead

krdln:
recheck no bug

PR #253, opened 2018-08-03
Silence the output of ReportUnit.exe

krdln:
To make our logs more readable


krdln:
> I'd prefer we save the output to a variable, and print it if last exit code is non zero.

This is the default behavior of `Invoke-NativeCommand`.

PR #255, opened 2018-08-07
Succeed build immediately when started from Zuul and other branch than master

sodar:


Jenkinsfile
```diff
@@ -1,6 +1,14 @@
 #!groovy
 library "contrailWindows@$BRANCH_NAME"
 
+// NOTE: This is a temporary workaround. Alternative workaround would be to put `when` in each stage,
```
m-kostrzewa: workaround for what?

PR #263, opened 2018-08-13
Get rid of utility/ subdir.

m-kostrzewa:
Split scripts into CI utility scripts underlk CIScripts/LogserverUtils.
Create a separate dir for manual CI scripts for CI ops (infra/).
This includes moving of update-zuul-layout to infra/.

TO REVIEWERS: do you agree with this change?


krdln:
A few thoughts:

1. I've just realized we have `CIScripts` directory in `ci` repository. This doesn't really make sense. Perhaps we can strip this layer?
2. If we stick to "all CI scripts should be in `CIScripts`", we should also move toplevel `Jenkinsfile` there (to avoid breaking Jenkins, we can temporarily create a symlink in the old location).
3. I disagree with moving `update-zuul-layout` out of ansible directory. This script is still tightly coupled to that directory, but now additionaly needs a weird readme and comments. I see a few alternatives (sorted from least to most preferable in my opinion):
   1. Make the script automatically find the location of `ansible` directory.
   2. Make the `ansible` directory or the `layout` file a mandatory parameter for the script.
   3. Leave the script in its original place for more code-data locality (and since I guess we want `infra` to contain *all* the infra-related scripts, we can add a thin wrapper there)
   4. (out of scope of this PR, I guess) Never invoke this script manually, but rather make it a part of zuul ansible role.

m-kostrzewa:
> I've just realized we have CIScripts directory in ci repository. This doesn't really make sense. Perhaps we can strip this layer?

It seems hard to do atm unfortunately, this repo is referenced in multiple places.

> If we stick to "all CI scripts should be in CIScripts", we should also move toplevel Jenkinsfile there (to avoid breaking Jenkins, we can temporarily create a symlink in the old location).

I'm OK with Jenkinsfiles being in another dir - in principle, some other CI system can also use scripts located in CIScripts repository. But no CIScript can call/import a Jenkinsfile.

> I disagree with moving update-zuul-layout out of ansible directory. This script is still tightly coupled to that directory, but now additionaly needs a weird readme and comments. I see a few alternatives (sorted from least to most preferable in my opinion):

> (out of scope of this PR, I guess) Never invoke this script manually, but rather make it a part of zuul ansible role.

TBH it doesn't seem that hard to do...

m-kostrzewa:
recheck no bug

m-kostrzewa:
recheck no bug

infra/update_zuul_layout/update-zuul-layout.py
```diff
@@ -5,6 +5,7 @@
 import sys
 import re
 
+# TODO: This script assumes that it's being ran from inside ansible/ directory.
 layout_file = "roles/zuul/files/layout.yaml"
```
jacek-izykowski: It seems that value of layout_file is the only thing that should be updated. We can probably avoid some problems (and questions) if we change it together with moving the script.

PR #265, opened 2018-08-14
Added defaults for zuul cloner

michal-clapinski:
those are copied from job to job in Jenkins, even though they are really useless there

### DIFF
--- a/CIScripts/Checkout.ps1
+++ b/CIScripts/Checkout.ps1
@@ -4,10 +4,28 @@
 
 $Job = [Job]::new("Checkout")
 
-Get-ZuulRepos -GerritUrl $Env:GERRIT_URL `
-              -ZuulProject $Env:ZUUL_PROJECT `
-              -ZuulRef $Env:ZUUL_REF `
-              -ZuulUrl $Env:ZUUL_URL `
-              -ZuulBranch $Env:ZUUL_BRANCH
+$defaults = @{
+    GERRIT_URL = "https://review.opencontrail.org"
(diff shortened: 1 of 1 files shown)


CIScripts/Checkout/Zuul.ps1
```diff
@@ -3,23 +3,21 @@
...
     )
 
     $ZuulClonerOptions = @(
+        "--zuul-branch=$($ZuulBranch)",
         "--map=./CIScripts/clonemap.yml",
         $GerritUrl
     )
 
-    if ($ZuulUrl -ne "") {
+    if ($ZuulAdditionalParams.Count -ne 0) {
```
krdln: Every type except basic ones (`string`, `integer`, `bool` etc.) can be null, and _will_ be null in case of non-present optional parameter. So it's best to check `if ($ZuulAdditionalParams -ne $null)` or just even `if ($ZuulAdditionalParams)`.

You could also make a class for this, so it's impossible to pass invalid hashtable, but meh.

CIScripts/Checkout.ps1
```diff
@@ -5,13 +5,17 @@
 $Job = [Job]::new("Checkout")
 
 if (Test-Path "Env:ZUUL_URL") {
+    $ZuulAdditionalParams = @{
+        Url = $Env:ZUUL_URL;
+        Project = $Env:ZUUL_PROJECT;
```
krdln: Sidenote nit: semicolons are unnecessary when newlines are present.

PR #266, opened 2018-08-14
Handle UPLOAD_ARTIFACTS in CI

michal-clapinski:
As you can see [here](http://10.84.12.26:8080/job/WinContrail/job/ci-contrail-windows-nightly-debug/configure) I specify `UPLOAD_ARTIFACTS=1`, but since [this job](http://10.84.12.26:8080/job/WinContrail/job/winci-server2016-prod) doesn't have this field, it's not passed there. For proof check [the Jenkins logs](http://10.84.12.26:8080/log/all):
```
Skipped parameter `UPLOAD_ARTIFACTS` as it is undefined on `WinContrail/winci-server2016-prod`.
```
Because of that reason I have to add the field `UPLOAD_ARTIFACTS` to winci-server2016-prod with default set to `0` and then set it to `1` in nightly.

### DIFF
--- a/CIScripts/BuildStage.ps1
+++ b/CIScripts/BuildStage.ps1
@@ -45,7 +45,7 @@ if ($NothingToBuild -or $CopyDisabledArtifacts) {
 }
 
 if (Test-Path Env:UPLOAD_ARTIFACTS) {
-    if ($Env:UPLOAD_ARTIFACTS -eq "yes") {
+    if ($Env:UPLOAD_ARTIFACTS -ne "0") {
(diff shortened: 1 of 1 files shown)


michal-clapinski:
recheck no bug

krdln:
I think the standard practice for boolean env variables is:

* no env variable → false
* ENV="" → false
* ENV=0 → false
* antything else → true

(but I don't feel strongly about it)

michal-clapinski:
recheck no bug

PR #268, opened 2018-08-17
UDP tests: Make UDP listener loop continue after an error

krdln:


Test/Tests/Protocol/TunnellingWithAgent.Tests.ps1
```diff
@@ -174,7 +174,7 @@ function Start-UDPEchoServerInContainer {
     '        $RemoteIPEndpoint.Port = $SendPort;' + `
     '        $UDPSocket.Send($Payload, $Payload.Length, $RemoteIPEndpoint);' + `
     '        \"Received message and sent it to: $RemoteIPEndpoint.\" | Out-String;' + `
-    '    }} catch {{ break }}' + `
+    '    }} catch {{ continue }}' + `
```
m-kostrzewa: I'm worried about debuggability of any UDP failures in the future.

At least it used to fail fast. Proposed change will cause a infinite loop.

Maybe we should log something here or have a max number of retries and then print last error or something?

PR #269, opened 2018-08-17
Retry loop for getting container IP

krdln:
_Perhaps_ we try to obtain the IP before it's set on the container. So let's add some logging and check if that hypothesis is true. I think that's impossible, but it's better to be sure.


krdln:
recheck no bug

krdln:
+kustosz @jablonskim

PR #274, opened 2018-08-21
Flake: failure talking to yum.

m-kostrzewa:
It seems like network flake.


m-kostrzewa:
It may not be a flake after all, it seems like a genuine failure in yum repo

PR #273, opened 2018-08-21
do not merge - test if golang bump worked

m-kostrzewa:

PR #275, opened 2018-08-21
Bump golang to 1.10.0.

m-kostrzewa:
This is because upgraded docker driver dependencies (hcsshim and
go-winio) rely on type aliasing mechanism, introduced in 1.9.0. But we
might as well skip to 1.10.0.


m-kostrzewa:
Test on single node: https://github.com/Juniper/contrail-windows-ci/pull/273

m-kostrzewa:
recheck no bug

PR #276, opened 2018-08-22
Support for release mode in nightly builds

aethyrxeo:
Divide nightly artifacts built in release or debug mode by moving them
into different folders.

### DIFF
--- a/CIScripts/BuildStage.ps1
+++ b/CIScripts/BuildStage.ps1
@@ -9,8 +9,11 @@ $OutputRootDirectory = "output"
 $NothingToBuild = $Env:COMPONENTS_TO_BUILD -eq "None"
 $CopyDisabledArtifacts = Test-Path Env:COPY_DISABLED_ARTIFACTS
 
+$IsReleaseMode = [bool]::Parse($Env:BUILD_IN_RELEASE_MODE)
+$BuildMode = $(if ($IsReleaseMode) { "production" } else { "debug" })
+
 if (-not $NothingToBuild) {
-    & $PSScriptRoot\Build.ps1
+    & $PSScriptRoot\Build.ps1 -BuildMode $BuildMode
 }
 
 if ($NothingToBuild -or $CopyDisabledArtifacts) {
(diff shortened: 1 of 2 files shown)


CIScripts/BuildStage.ps1
```diff
@@ -9,8 +9,11 @@ $OutputRootDirectory = "output"
 $NothingToBuild = $Env:COMPONENTS_TO_BUILD -eq "None"
 $CopyDisabledArtifacts = Test-Path Env:COPY_DISABLED_ARTIFACTS
 
+$IsReleaseMode = [bool]::Parse($Env:BUILD_IN_RELEASE_MODE)
+$BuildMode = $(if ($IsReleaseMode) { "production" } else { "debug" })
+
 if (-not $NothingToBuild) {
-    & $PSScriptRoot\Build.ps1
+    & $PSScriptRoot\Build.ps1 -BuildMode $BuildMode
```
m-kostrzewa: Thought: it's not entirely clear from $BuildMode variable name what it means. This seems to "leaks" from internal implementation details.

aethyrxeo: I tried to hide the implementation details by adding a helper function, but maybe it's not an elegant solution. What do you think?

CIScripts/BuildStage.ps1
```diff
@@ -47,7 +48,8 @@ if ($NothingToBuild -or $CopyDisabledArtifacts) {
 if (Test-Path Env:UPLOAD_ARTIFACTS) {
     if ($Env:UPLOAD_ARTIFACTS -ne "0") {
         $ArtifactsPath = "\\$Env:SHARED_DRIVE_IP\SharedFiles\WindowsCI-UploadedArtifacts"
-        $Subdir = "$Env:JOB_NAME\$Env:BUILD_NUMBER"
+        $NightlyBuildMode = Resolve-BuildMode
```
michal-clapinski: it's not only for nightly, so change the name

Jenkinsfile
```diff
@@ -150,6 +149,12 @@ pipeline {
                         unstash "CIScripts"
                         unstash "SourceCode"
 
+                        script {
```
michal-clapinski: let's keep Jenkinsfile as simple as possible, just move it to powershell

PR #278, opened 2018-08-23
Add simple heuristic for fault detection.

m-kostrzewa:
This should add a prefix to split log file that we suspect contains the error, based on Jenkins output when a stage is failed. This should make it easier to look for errors in logs.

### DIFF
--- a/utility/split-log-into-stages.sh
+++ b/utility/split-log-into-stages.sh
@@ -29,7 +29,7 @@ do
     echo "copying '$stage' log to $stage_log_filename"
     grep --perl-regexp "$this_stage_regexp" "$full_log" | gzip > "$stage_log_filename"
 
-    if grep "Failed in branch" $stage_log_filename
+    if zgrep "Failed in branch" $stage_log_filename
     then
         mv $stage_log_filename "SUSPECTED-ERROR-HERE-$stage_log_filename"
     fi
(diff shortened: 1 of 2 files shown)


utility/split-log-into-stages.sh
```diff
@@ -28,6 +28,11 @@ do
     this_stage_regexp=$(get_stage_regexp "$stage")
     echo "copying '$stage' log to $stage_log_filename"
     grep --perl-regexp "$this_stage_regexp" "$full_log" | gzip > "$stage_log_filename"
+
+    if grep "Failed in branch" temp.txt
```
krdln: What is temp.txt?

utility/split-log-into-stages.sh
```diff
@@ -28,6 +28,11 @@ do
...
+    then
+        mv $stage_log_filename "SUSPECTED-ERROR-HERE-$stage_log_filename"
```
krdln: I'd prefer the log name to be independent of failure occurence and have some additional file that describes the failure.

m-kostrzewa:
pls fail

m-kostrzewa:
recheck windows

utility/split-log-into-stages.sh
```diff
@@ -29,7 +29,7 @@ do
     echo "copying '$stage' log to $stage_log_filename"
     grep --perl-regexp "$this_stage_regexp" "$full_log" | gzip > "$stage_log_filename"
 
-    if grep "Failed in branch" temp.txt
+    if grep "Failed in branch" $stage_log_filename
```
krdln: This is a `.gz` file. Please use `zgrep` or `zcat | grep`.

m-kostrzewa: TIL

m-kostrzewa:
recheck windows

krdln:
recheck no bug

PR #285, opened 2018-08-28
Verify tunneling method with vrfstats.

m-kostrzewa:


m-kostrzewa:
recheck no bug

sodar:
recheck no bug

PR #297, opened 2018-09-05
 Allow PSScriptAnalyzer installed by choco in Invoke script.

m-kostrzewa:
We used to install PSScriptAnalyzer using PSget. So we checked its
version using Get-Package. But now it's installed by chocolatey, so it
doesn't appear in Get-Package. Instead, we need to check Get-Module.

### DIFF
--- a/StaticAnalysis/Invoke-StaticAnalysisTools.ps1
+++ b/StaticAnalysis/Invoke-StaticAnalysisTools.ps1
@@ -11,12 +11,15 @@ function Assert-MeetsMinimalAnalyzerVersion($MinimumVersion) {
         $Analyzer = Get-Package -Name PSScriptAnalyzer
     } catch {
         # Maybe it was installed by chocolatey?
+        try {
+            Import-Module -Name PSScriptAnalyzer
+        } catch {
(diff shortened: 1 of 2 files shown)


StaticAnalysis/Invoke-StaticAnalysisTools.ps1
```diff
@@ -1,32 +1,49 @@
...
-} catch {
-    Write-Host "PSScriptAnalyzer not found. If you have installed it, make sure that it's at least version $MinimumVersion."
-    exit 1
+function Assert-MeetsMinimalAnalyzerVersion($MinimumVersion) {
+    try {
+        # Maybe it was installed using PSGet?
+        $Analyzer = Get-Package -Name PSScriptAnalyzer
+    } catch {
+        # Maybe it was installed by chocolatey?
+        $Analyzer = Get-Module -Name PSScriptAnalyzer
```
sodar: Does `Get-Module` throw an exception?

m-kostrzewa:
recheck no bug

m-kostrzewa:
recheck no bug

PR #298, opened 2018-09-07
Run flake detection before sending logs.

m-kostrzewa:
This way we can also see log output for flake detection scripts.


m-kostrzewa:
recheck no bug

PR #302, opened 2018-09-12
jenkins: Use builder label in Static analysis

sodar:
Since PSScriptAnalyser is already deployed on builder nodes, `builder-choco-analyzer` label can removed.


sodar:
recheck no bug

sodar:
recheck no bug

PR #303, opened 2018-09-12
Fixes using newest Shelly.

m-kostrzewa:


Test/Utils/CommonTestCode.ps1
```diff
@@ -1,3 +1,4 @@
+. $PSScriptRoot\..\..\CIScripts\Common\Init.ps1
```
m-kostrzewa: Changes like this involve explicitely importing files that we use, instead of relying on order of loading them in runtime

Test/Tests/Integration/NodeManager.Tests.ps1
```diff
@@ -85,7 +85,7 @@ function Start-NodeMgr {
         Justification="Analyzer doesn't understand relation of Pester blocks"
         )]
         $NodeMgrJob = Start-Job -ScriptBlock {
-            contrail-nodemgr --nodetype contrail-vrouter
+            contrail-nodemgr --nodetype contrail-vrouter # allow unknown-functions(contrail-nodemgr)
```
krdln: I think it's better to add `contrail-nodemgr` to `extras` in `shelly.toml` to leave the code cleaner

backups/Backup-Infrastructure.ps1
```diff
@@ -1,3 +1,5 @@
+. $PSScriptRoot/../CIScripts/Common/Init.ps1
+
```
krdln: Are we sure enabling strict mode in these backup scripts won't break them?

Test/Utils/RegisterDefaultResorces.ps1
```diff
@@ -1,5 +1,7 @@
 . $PSScriptRoot\..\..\CIScripts\Common\Init.ps1
+. $PSScriptRoot\..\..\CIScripts\Common\Init.ps1
```
krdln: huh?

Test/Utils/CommonTestCode.ps1
krdln: I'm not sure I get this comment.

Previously, we imported `Init.ps1` only in the top-level "script" PowerShell files. This file is a "library" PowerShell file. Do we want to do it that way?

m-kostrzewa:
Reopening in https://github.com/Juniper/contrail-windows-ci/pull/334

PR #304, opened 2018-09-12
c-a-d: Limit Java memory usage

sodar:
Added Java options to limit a maximum allocation pool to 2 GB, to avoid _hungry Cassandra_ syndrome.


sodar:
recheck no bug

sodar:
recheck no bug

michal-clapinski:
what about the other 2 options?

PR #305, opened 2018-09-13
zuulv3 tentacle and nightly containers

michal-clapinski:
still kinda WIP


CIScripts/Checkout/Tentacle.ps1
```diff
@@ -0,0 +1,13 @@
+function Get-TentacleRepos {
```
m-kostrzewa: We probably want to add a longer comment here explaining what the tentacle is

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -401,3 +401,58 @@ function Invoke-AgentTestsBuild {
...
+
+    $Job.PushStep("Containers build")
+
+    $Job.Step("add insecure registry", {
+        $DockerConfig = @"
+{
+    "insecure-registries" : [ "$Registry" ]
+}
+"@
+        Set-Content -Path "C:\ProgramData\Docker\config\daemon.json" -Value $DockerConfig
```
m-kostrzewa: ehh, so what are the ideas we have for this?

sodar: We can limit a number of executors on each builder to 1.

sodar: Limiting executors to 1 removes _docker race_ problem in the future. To achieve the same build throughput, we can simply scale the number of builds.

m-kostrzewa: Since we (codi) are the ones that setup the insecure registries, can we possibly just use constant number of possible port numbers, like a 100 or something? Then we could generate daemon.json file once with 100 entries.

michal-clapinski: That would lock us to current IP address of the registry machine. Let's go with limiting the number of executors.

CIScripts/Checkout/Tentacle.ps1
```diff
@@ -0,0 +1,16 @@
+# Tentacle is a component for running Jenkins jobs from Zuul v3
+# It serves an HTTP server with repositories. We need to download and upack them.
```
sodar: `upack` -> `unpack`

CIScripts/Build.ps1
```diff
@@ -84,6 +86,29 @@ try {
...
+                    $NodemgrOutputDir
+                )
+            },
+            @{
+                Suffix = "docker-driver"
+                Folders = @(
+                    $DockerDriverOutputDir
+                )
+            }
+        )
```
sodar: Instead of using a hashtable we could define a class `ContainerAttribute`, e.g. like this:

```powershell
class ContainerAttributes {
    [string] $Suffix;
    [string[]] $Folders;

    ContainerAttributes([string] $Suffix, [string[]] $Folders) {
        $this.Suffix = $Suffix;
        $this.Folders= $Folders;
    }
}
```

Array construction would look like this:

```powershell
$ContainersAttributes = @(
    [ContainerAttributes]::New("vrouter", @(
        $vRouterOutputDir,
        $AgentOutputDir,
        $NodemgrOutputDir
    )),
    [ContainerAttributes]::New("docker-driver", @(
        $DockerDriverOutputDir
    ))
)
```

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -401,3 +401,61 @@ function Invoke-AgentTestsBuild {
 
     $Job.PopStep()
 }
+
+function Invoke-ContainerBuild {
```
sodar: We could move `Invoke-ContainerBuild` and `Invoke-ContainersBuild` to a new file named ` CIScripts/Build/Container.ps1`.

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -401,3 +401,61 @@ function Invoke-AgentTestsBuild {
...
+    "insecure-registries" : [ "$Registry" ]
+}
+"@
+        Set-Content -Path "C:\ProgramData\Docker\config\daemon.json" -Value $DockerConfig
+        Restart-Service docker
+    })
+
+    ForEach ($ContainerAttributes in $ContainersAttributes) {
+        $Job.Step("Building contrail-windows-$($ContainerAttributes.Suffix)", {
+            Invoke-ContainerBuild -WorkDir $WorkDir `
```
sodar: Maybe we should pass `$ContainerAttributes` object directly to `Invoke-ContainerBuild`, instead of using two separate parameters for suffix and folders list.

CIScripts/BuildStage.ps1
```diff
@@ -17,7 +17,7 @@ if (-not $NothingToBuild) {
 if ($NothingToBuild -or $CopyDisabledArtifacts) {
     $Job = [Job]::new("Copying ready artifacts")
 
-    $ArtifactsPath = "\\$Env:SHARED_DRIVE_IP\SharedFiles\WindowsCI-Artifacts"
+    $ArtifactsPath = "\\$Env:SHARED_DRIVE_IP\SharedFiles\WindowsCI-UploadedArtifacts\WinContrail\winci-server2016-devel\debug\3385"
```
sodar: Is this some leftover from testing?

PR #308, opened 2018-09-17
Move Test- methods to separate directory.

m-kostrzewa:
Test suite files were getting too cluttered and there was some duplication with Test-Ping.

(Not only it is used between two files, but also move tests that would use it are coming soon - DNS).


m-kostrzewa:
Depends on

m-kostrzewa:
recheck no bug

m-kostrzewa:
recheck no bug

m-kostrzewa:
recheck no bug

m-kostrzewa:
recheck no bug

m-kostrzewa:
recheck no bug

PR #312, opened 2018-09-24
Use expand-archive from powershell instead of pscx

michal-clapinski:
We removed PSCX so we have to change the usage.


michal-clapinski:
recheck no bug

PR #316, opened 2018-09-25
Compute diagnostic check script - partial implementation.

m-kostrzewa:


sodar:
recheck no bug

m-kostrzewa:
@jacekchalupka good point

m-kostrzewa:
recheck no bug

PR #319, opened 2018-09-26
Fix error while cleaning up NAT and Container networks when default N…

jablonskim:
…AT network is inaccessible

### DIFF
--- a/Test/TestConfigurationUtils.ps1
+++ b/Test/TestConfigurationUtils.ps1
@@ -166,10 +166,14 @@ function Stop-DockerDriver {
     Invoke-Command -Session $Session -ScriptBlock {
         Stop-Service docker | Out-Null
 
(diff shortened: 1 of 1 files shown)


Test/TestConfigurationUtils.ps1
```diff
@@ -165,9 +165,14 @@ function Stop-DockerDriver {
 
     Invoke-Command -Session $Session -ScriptBlock {
         Stop-Service docker | Out-Null
-        Get-NetNat | Remove-NetNat -Confirm:$false
-        Get-ContainerNetwork | Remove-ContainerNetwork -ErrorAction SilentlyContinue -Force
-        Get-ContainerNetwork | Remove-ContainerNetwork -Force
+
+        Get-NetNat | ForEach-Object {
```
sodar: We should put a comment describing why we remote NAT networks and container networks this way.

PR #320, opened 2018-09-26
Fix 'docker build' failure when nat network is inaccessible

jablonskim:


jablonskim:
recheck no bug

jablonskim:
recheck no bug

jablonskim:
recheck no bug

jablonskim:
recheck no bug

PR #321, opened 2018-09-26
Fix conflict while creating virtual network when previous test was no…

jablonskim:
…t finished properly


jablonskim:
recheck no bug

PR #323, opened 2018-09-27
don't run CI on recheck comment if already +1'd

michal-clapinski:


michal-clapinski:
recheck no bug

michal-clapinski:
recheck no bug

michal-clapinski:
recheck no bug

PR #326, opened 2018-09-28
Use Test-Ping instead of raw ping

michal-clapinski:
Because we have an awesome wrapper for `ping` in `Test-Ping`


m-kostrzewa:
:1st_place_medal:

PR #331, opened 2018-10-01
TestConfigurationUtils refactor for NSSM introduction

aethyrxeo:


Test/Utils/ComputeNode/Initialize.ps1
```diff
@@ -3,6 +3,162 @@
...
+
+    Invoke-Command -Session $Session -ScriptBlock {
+        New-ContainerNetwork -Mode Transparent -NetworkAdapterName $Using:AdapterName -Name $Using:ContainerNetworkName | Out-Null
+    }
+
+    # We're not waiting for IP on this adapter, because our tests
+    # don't rely on this adapter to have the correct IP set for correctess.
+    # We could implement retrying to avoid flakiness but it's easier to just
+    # ignore the error.
+    # Wait-RemoteInterfaceIP -Session $Session -AdapterName $SystemConfig.VHostName
```
m-kostrzewa: ded code? remove? discuss pls

aethyrxeo: [this](https://github.com/Juniper/contrail-windows-ci/pull/179) commit is responsible for the comment.

Idk rly what to do about it. I thought of refactor at "function-level" -> move logically similar functions to one file, to make TestConfigurationUtils filled with only "pure utility" functions and leave the function bodies intact.

m-kostrzewa: @krdln do you remember the reason for commenting out this piece of code? Can we remove it since we have implemented retries for tests? @jacek-izykowski what do you think?

krdln: So, usually after creating vEthernet adapter, we wait until the original IP from appears on the new vEtherent adapter. Sometimes it doesn't, in which case the adapter is semi-broken.

The Docker Driver handles it by retrying the adapter creation until it creates one with a valid IP. Here, we don't have such a retry loop, so when this line wasn't commented out, it caused flakiness (as sometimes the adapter didn't have IP). Ideally, similar retry loop could be added here. But it turns out that vtest tests work fine with semi-broken adapter without IP, so the line was commented out.

I don't know anything about retrying tests, so I won't comment on that.

m-kostrzewa: So I guess we can remove it?

jacek-izykowski: Let's remove this code.

aethyrxeo: Ok.

PR #335, opened 2018-10-02
Add more verbose msg when DockerDriver tests are missing

jablonskim:


Test/Tests/Component/DockerDriver.Tests.ps1
```diff
@@ -21,10 +21,18 @@ function Find-DockerDriverTests {
         [Parameter(Mandatory=$true)] [PSSessionT] $Session,
         [Parameter(Mandatory=$true)] [string] $RemoteSearchDir
     )
+
     $TestModules = Invoke-Command -Session $Session {
-        Get-ChildItem -Recurse -Filter "*.test.exe" -Path $Using:RemoteSearchDir `
-            | Select-Object BaseName, FullName
+        $Tests = @(Get-ChildItem -Recurse -Filter "*.test.exe" -Path $Using:RemoteSearchDir)
+        if ($Tests.Count -eq 0) {
```
m-kostrzewa: Can't we just 
```
$a = Get-ChildItem
if (-not $a) {...} 
```?

Comment (jablonskim) [reply Test/Tests/Component/DockerDriver.Tests.ps1]: Yes, it should work, but I prefer the array way. In my opinion it is better to expect empty array when no elements were found.

Comment (m-kostrzewa) [reply Test/Tests/Component/DockerDriver.Tests.ps1]: `Get-ChildItem` already returns an array, and you can check if it's empty using `-not`:
```
$a = @()
if(-not $a) { echo "this will be printed" }
```
You're adding this array unnecessarily. It makes the reader question why is it there, when in fact it's redundant code.

Comment (m-kostrzewa) [reply Test/Tests/Component/DockerDriver.Tests.ps1]:
```
mkdir tmp
cd tmp
$a = Get-ChildItem
if(-not $a) { echo "this will be printed as well" }
```

jablonskim: `Get-ChildItem` returns:
  - an array when there are two elements or more,
  - object, when there are single element,
  - `null`, when there are no elements

jablonskim:
```PS C:\Users\mj> $a = Get-ChildItem *
PS C:\Users\mj> $a.GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     Object[]                                 System.Array


PS C:\Users\mj> $a = Get-ChildItem .\a.out
PS C:\Users\mj> $a.GetType()

IsPublic IsSerial Name                                     BaseType
-------- -------- ----                                     --------
True     True     FileInfo                                 System.IO.FileSystemInfo


PS C:\Users\mj> $a = Get-ChildItem asdasd*
PS C:\Users\mj> $a.GetType()
You cannot call a method on a null-valued expression.
At line:1 char:1
+ $a.GetType()
+ ~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], RuntimeException
    + FullyQualifiedErrorId : InvokeMethodOnNull```

PR #339, opened 2018-10-03
 Install NSSM for service creation

aethyrxeo:
We do this in seperate commit, because we need to update testbeds.


aethyrxeo:
[create testbed job](http://10.84.12.26:8080/job/WinContrail-infra/job/templates/job/create-testbed-template/77/)

PR #340, opened 2018-10-03
Create Agent Service with NSSM

aethyrxeo:
After this and future changes in c-a-d are merged, agent's service wrapper is no longer necessary and can be removed.


Test/Utils/ComputeNode/Service.ps1
```diff
@@ -97,3 +104,81 @@ function Out-StdoutAndStderrToLogFile {
...
+    Param (
+        [Parameter(Mandatory=$true)] $Session
+    )
+
+    Write-Log "Starting Agent"
+
+    $AgentServiceName = Get-AgentServiceName
+    $Output = Invoke-NativeCommand -Session $Session -ScriptBlock {
+        $Output = netstat -abq  #dial tcp bug debug output
+        #TODO: After the bugfix, use Enable-Service generic function here.
```
m-kostrzewa: Which bugfix?

aethyrxeo: bugfix for `dial tcp 127.0.0.1:9091: connectex: No connection could be made because the target machine actively refused it.` bug

m-kostrzewa: OK so please write it in comment so that we know in the future

PR #330, opened 2018-10-01
Create CNM Plugin service with NSSM

aethyrxeo:

PR #344, opened 2018-10-10
Add functions for service utility

aethyrxeo:


Test/Utils/ComputeNode/Service.ps1
```diff
@@ -0,0 +1,93 @@
...
+
+    $NSSMServiceAlreadyCreatedError = 5
+    if ($Output.ExitCode -eq 0) {
+        Write-Log $Output.Output
+    }
+    elseif ($Output.ExitCode -eq $NSSMServiceAlreadyCreatedError) {
+        Write-Log "$ServiceName service already created, continuing..."
+    }
+    else {
+        throw [HardError]::new("Unknown (wild) error appeared while creating $ServiceName service")
```
sodar: Let's put `$Output` and exit code in the exception message.

Test/Utils/ComputeNode/Service.ps1
```diff
@@ -0,0 +1,93 @@
...
+    }
+}
+
+function Out-StdoutAndStderrToLogFile {
+    Param (
+        [Parameter(Mandatory=$true)] $Session,
+        [Parameter(Mandatory=$true)] $ServiceName,
+        [Parameter(Mandatory=$true)] $LogPath
+    )
+    #redirect stdout and stderr to the log file
```
sodar: This comment is not needed here. Function in self-explanatory.

Test/Utils/ComputeNode/Service.ps1
```diff
@@ -0,0 +1,93 @@
...
+    )
+
+    $Output = Invoke-NativeCommand -Session $Session {
+        nssm remove $using:ServiceName confirm
+    } -CaptureOutput
+
+    Write-Log $Output
+}
+
+function Enable-Service {
```
sodar: Enable/Disable is not a good name in my opinion. E.g. on systemd Enable/disable is used to signify if service is started automatically.

Maybe `Enable-Service` should be named `Start-RemoteService`. The same applies to `Disable-Service`.

aethyrxeo:
This functions are used in [Create CNM Plugin service with NSSM](https://github.com/Juniper/contrail-windows-ci/pull/330) and [Create Agent Service with NSSM](https://github.com/Juniper/contrail-windows-ci/pull/340)

Test/Utils/ComputeNode/Service.ps1
```diff
@@ -0,0 +1,99 @@
...
+}
+
+function Stop-RemoteService {
+    Param (
+        [Parameter(Mandatory=$true)] $Session,
+        [Parameter(Mandatory=$true)] $ServiceName
+    )
+
+    Invoke-Command -Session $Session -ScriptBlock {
+        Stop-Service $using:ServiceName -ErrorAction SilentlyContinue
```
m-kostrzewa: Why is it ok to `ErrorAction SilentlyContinue` but not in `Start-RemoteService`?

Test/Utils/ComputeNode/Service.ps1
```diff
@@ -0,0 +1,99 @@
...
+function Out-StdoutAndStderrToLogFile {
+    Param (
+        [Parameter(Mandatory=$true)] $Session,
+        [Parameter(Mandatory=$true)] $ServiceName,
+        [Parameter(Mandatory=$true)] $LogPath
+    )
+
+    $Output = Invoke-NativeCommand -Session $Session -ScriptBlock {
+        nssm set $using:ServiceName AppStdout $using:LogPath
+        nssm set $using:ServiceName AppStderr $using:LogPath
```
jablonskim: Does NSSM handle this situation (stdout and stderr redirected to single file) correctly?

aethyrxeo: From what i saw, yes.

However, there is a case when deleting service and creating new one, that the old log file will be overwritten.

PR #345, opened 2018-10-10
 Contrail API DNS functions

jacekchalupka:
Adding/removing DNS servers
Adding/removing DNS records
Setting DNS mode in IPAM


Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
```diff
@@ -0,0 +1,132 @@
...
+            -Username $OpenStackConfig.Username `
+            -Password $OpenStackConfig.Password `
+            -Tenant $OpenStackConfig.Project
+    }
+
+    It 'adds and removes "empty" contrail DNS server' {
+        $ServerUUID = Add-ContrailDNSServer -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+            -DNSServerName "CreatedByPS1Script"
+        Remove-ContrailDNSServer -ContrailUrl $ContrailUrl -AuthToken $AuthToken -DNSServerUuid $ServerUUID
+    }
```
m-kostrzewa: This test (and some others) don't assert anything. If we expect that something should not throw, then we should probably use `Should -Not -Throw` assert.

Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
```diff
@@ -0,0 +1,132 @@
...
+        $Record2UUID = Add-ContrailDNSRecord -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+            -DNSServerName "CreatedByPS1Script" -VirtualDNSRecordData $Record2Data
+
+        Remove-ContrailDNSRecord -ContrailUrl $ContrailUrl -AuthToken $AuthToken -DNSRecordUuid $Record1UUID
+        Remove-ContrailDNSRecord -ContrailUrl $ContrailUrl -AuthToken $AuthToken -DNSRecordUuid $Record2UUID
+
+        Remove-ContrailDNSServer -ContrailUrl $ContrailUrl -AuthToken $AuthToken -DNSServerUuid $ServerUUID
+    }
+
+    It 'breaks when wrong dns mode specified' {
```
m-kostrzewa: s/breaks/throws

Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
```diff
@@ -0,0 +1,132 @@
...
+        { Set-IpamDNSMode -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+            -IpamFQName @("default-domain", "default-project", "default-network-ipam") `
+            -DNSMode 'wrongdnsmode' } | Should -Throw
+    }
+
+    It 'sets DNS mode to none' {
+        Set-IpamDNSMode -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
```
m-kostrzewa: Maybe we should create `Context` blocks, like:

* Context "DNS server creation and removal"
* Context "Setting DNS modes"

Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
```diff
@@ -0,0 +1,132 @@
...
+            -DNSMode 'tenant-dns-server'
+    }
+
+    It 'sets DNS mode to tenant with DNS servers' {
+        Set-IpamDNSMode -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+            -IpamFQName @("default-domain", "default-project", "default-network-ipam") `
+            -DNSMode 'tenant-dns-server' -TenantServersIPAddresses @("1.1.1.1", "2.2.2.2")
+    }
+
+    It 'sets DNS mode to virtual' {
```
m-kostrzewa: Test names like this one suggest that it always does set something. We should probably name them `It 'can set DNS mode to virtual` and similar in other places.

Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
```diff
@@ -0,0 +1,132 @@
...
+        $ServerUUID = Add-ContrailDNSServer -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+            -DNSServerName "CreatedByPS1Script"
+
+        Set-IpamDNSMode -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+            -IpamFQName @("default-domain", "default-project", "default-network-ipam") `
+            -DNSMode 'virtual-dns-server' -VirtualServerName "CreatedByPS1Script"
+
+        Remove-ContrailDNSServer -ContrailUrl $ContrailUrl -AuthToken $AuthToken -DNSServerUuid $ServerUUID -Force
```
m-kostrzewa: Why do we `Add` and `Remove` `ContrailDNSServer` in this test case, but not in other which verify `Set-IpamDNSMode` function?

m-kostrzewa: This hints at lack of test case isolation http://wiki.c2.com/?UnitTestIsolation

Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
m-kostrzewa: We can further split "Setting DNS modes" by DNS mode types maybe?

Test/Utils/ContrailAPI/ConfigureDNS.ps1
```diff
@@ -0,0 +1,174 @@
...
+        }
+        $Request."network-ipam"."virtual_DNS_refs" = @($DNSServerRef)
+    }
+
+    $RequestUrl = $ContrailUrl + "/network-ipam/" + $IpamUuid
+    $Response = Invoke-RestMethod -Uri $RequestUrl -Headers @{"X-Auth-Token" = $AuthToken} `
+        -Method Put -ContentType "application/json" -Body (ConvertTo-Json -Depth $CONVERT_TO_JSON_MAX_DEPTH $Request)
+
+    return $Response.'network-ipam'.'uuid'
+}
```
m-kostrzewa: This function is pretty long, can we split it to smaller functions? Especially, since we modify state of a single struct throughout the whole body. It makes it hard to track.

Test/Utils/ContrailAPI/ConfigureDNS.ps1
```diff
@@ -0,0 +1,174 @@
...
+        [Parameter(Mandatory = $true)] [string] $AuthToken,
+        [Parameter(Mandatory = $true)] [string[]] $IpamFQName,
+        [Parameter(Mandatory = $true)]
+            [ValidateSet("none","default-dns-server","tenant-dns-server","virtual-dns-server")]
+            [string] $DNSMode,
+        [Parameter(Mandatory = $false)] [string[]] $TenantServersIPAddresses = @(),
+        [Parameter(Mandatory = $false)] [string] $VirtualServerName
+    )
+
+    $IpamUuid = FQNameToUuid -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
```
m-kostrzewa: This variable is not used until much later, maybe let's move it closer to where it's used.

Test/Utils/ContrailAPI/ConfigureDNS.ps1
m-kostrzewa: Maybe something like:
```
Set-IpamDNSMode {
   $Request = {...}
   if (tenant) { $Request =  AddTenantStuff($Request) }
   elif (virtual) { $Request = AddVirtualStuff($Request) }
   ...
}
```

Test/Utils/ContrailAPI/ConfigureDNS.ps1
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    $RequestUrl = $ContrailUrl + "/network-ipam/" + $IpamUuid
+    $Response = Invoke-RestMethod -Uri $RequestUrl -Headers @{"X-Auth-Token" = $AuthToken} `
+        -Method Put -ContentType "application/json" -Body (ConvertTo-Json -Depth $CONVERT_TO_JSON_MAX_DEPTH $Request)
+
+    return $Response.'network-ipam'.'uuid'
+}
+
+function Add-ContrailDNSRecord {
```
m-kostrzewa: [pedantic] According to https://docs.microsoft.com/en-us/powershell/developer/cmdlet/approved-verbs-for-windows-powershell-commands, `Add` verb:

```
Adds a resource to a container, or attaches an item to another item. For example, the Add-Content cmdlet adds content to a file. This verb is paired with Remove.
```

How about using `New`?

Test/Utils/ContrailAPI/ConfigureDNS.ps1
```diff
@@ -0,0 +1,174 @@
...
+        $NetworkIPAMs = $DNSServer.'virtual-DNS'.'network_ipam_back_refs'
+        ForEach ($NetworkIPAM in $NetworkIPAMs) {
+            Set-IpamDNSMode -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+                -IpamFQName $NetworkIPAM.'to' `
+                -DNSMode 'none'
+        }
+    }
+
+    Invoke-RestMethod -Method Delete -Uri $DNSServerURl -Headers @{"X-Auth-Token" = $AuthToken} | Out-Null
+}
```
m-kostrzewa: BTW, have you tried generating test coverage report for this file? If you add CI tag to the Describe block, you can call `Invoke-Selfcheck` with `-CodeCoverage` flag - then you can grep for your files and check what is missing and whether you want to do something about it. 

BTW2 just learned that VScode can't do it in the editor window: https://github.com/PowerShell/vscode-powershell/issues/495 :(

Test/Utils/ContrailAPI/Global.ps1
```diff
@@ -0,0 +1,16 @@
+function FQNameToUuid {
```
m-kostrzewa: This function is used in only one file, therefore why is it in "Global.ps1" file? It's like the most local thing I've ever seen :)

We should probably move it to file where it's used and think about extracting it to separate file when another client of this function is needed OR rename this file to `FQName.ps1` or similar.

Test/Utils/ContrailAPI/Global.ps1
```diff
@@ -0,0 +1,16 @@
...
+           [Parameter(Mandatory = $true)] [string[]] $FQName)
+
+    $Request = @{
+        type     = $Type
+        fq_name  = $FQName
+    }
+
+    $FQToUuidUrl = $ContrailUrl + "/fqname-to-id"
+    $Response = Invoke-RestMethod -Uri $FQToUuidUrl -Headers @{"X-Auth-Token" = $AuthToken} `
+        -Method Post -ContentType "application/json" -Body (ConvertTo-Json -Depth $CONVERT_TO_JSON_MAX_DEPTH $Request)
```
m-kostrzewa: https://github.com/krdln/shelly would complain that `$CONVERT_TO_JSON_MAX_DEPTH` is an indirect import. Not that important, but just wanted to plug this awesome tool :)

Test/Utils/ContrailAPI/VirtualDNSData.ps1
```diff
@@ -0,0 +1,38 @@
+class VirtualDNSData {
+    [string] $DomainName;
+    [boolean] $DynamicRecordsFromClient;
+    [int] $DefaultTTL; # In seconds
```
m-kostrzewa: Then call it `$DefaultTTLInSeconds` and remove the comment

Test/Utils/ContrailAPI/VirtualDNSData.ps1
```diff
@@ -0,0 +1,38 @@
...
+    [boolean] $ExternalVisible;
+    [boolean] $ReverseResolution;
+
+    [ValidateSet("fixed","random","round-robin")]
+    [string] $RecordOrder;
+
+    [ValidateSet("dashed-ip","dashed-ip-tenant-name","vm-name","vm-name-tenant-name")]
+    [string] $FloatingIpRecord;
+
+    hidden Init([boolean] $DynamicRecordsFromClient, [string] $RecordOrder,
```
m-kostrzewa: TIL about `hidden` keyword, pretty cool.

Test/Utils/ContrailAPI/VirtualDNSData.ps1
```diff
@@ -0,0 +1,38 @@
...
+    [int] $DefaultTTL, [string] $FloatingIpRecord,
+    [boolean] $ExternalVisible, [boolean] $ReverseResolution) {
```
m-kostrzewa: [pedantic] Hmm, maybe let's add one more level of indent for parameters here?

Test/Utils/ContrailAPI/VirtualDNSRecordData.ps1
```diff
@@ -0,0 +1,42 @@
+class VirtualDNSRecordData {
+    [string] $RecordName;
+    [string] $RecordClass;
+    [string] $RecordData;
+    [int] $RecordTTL; # In seconds
```
m-kostrzewa: Same here as above ^

Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
jacekchalupka: Because other modes doesn't use Contrail DNS Servers. This one does so we need to create one.

Test/Utils/ContrailAPI/ConfigureDNS.ps1
jacekchalupka: This name i consistent with other ContrailAPI functions we use in this repository. And actually it adds resource to container. It adds new DNS Record to DNS Server (creating it in a process, but we cannot just create a record).

Test/Utils/ContrailAPI/Global.ps1
jacekchalupka: It is global from the API level. It can be used for all the resources.  If somebody will be writing another API fragment, he will not be searching all the files to check if somebody wrote some functions like this. Maybe we can name this file "UniversalFunctions" or something like that?

Test/Utils/ContrailAPI/Global.ps1
jacekchalupka: That is strange, because I directly imported it at the the first line of file. Maybe it is because another imported file also imports it and overrides the value.

Test/Utils/ContrailAPI/ConfigureDNS.ps1
jacekchalupka: Nearly 100%, two constructors are not covered, but they just invoke init method.

Test/Utils/ContrailAPI/ConfigureDNS.Tests.ps1
```diff
@@ -0,0 +1,167 @@
...
+                $ServerUUID = Add-ContrailDNSServer -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+                    -DNSServerName "CreatedByPS1ScriptForIpam"
+
+                {
+                    Set-IpamDNSMode -ContrailUrl $ContrailUrl -AuthToken $AuthToken `
+                        -IpamFQName @("default-domain", "default-project", "default-network-ipam") `
+                        -DNSMode 'virtual-dns-server' -VirtualServerName "CreatedByPS1ScriptForIpam"
+                } | Should -Not -Throw
+
+                Remove-ContrailDNSServer -ContrailUrl $ContrailUrl -AuthToken $AuthToken -DNSServerUuid $ServerUUID -Force
```
m-kostrzewa: If assert fails, this tear down step won't be called. Either use BeforeEach/AfterEach combo, or try..finally.

Test/Utils/ContrailAPI/Global.ps1
m-kostrzewa: Ooooh, so that's why it's called Globals.ps1. If that's the reason then it's fine.

jacekchalupka:
recheck no bug

PR #363, opened 2018-10-22
Added DNS tests

jacekchalupka:


Test/Tests/Network/DNS.Tests.ps1
```diff
@@ -0,0 +1,386 @@
...
+                    | Should -BeTrue
+            }
+        }
+
+        Context "DNS mode virtual" {
+            BeforeAll { BeforeEachContext -DNSSetting ([VirtualDNSSettings]::New($VirtualDNSServer.GetFQName())) }
+
+            AfterAll { AfterEachContext }
+
+            It "resolves juniper.net" {
```
aethyrxeo: Shouldn't we add tag for these tests so they are only executed in CI?

jacekchalupka:
http://10.84.12.26:8080/job/WinContrail-infra/job/templates/job/create-testbed-template/78/

Test/Tests/Network/DNS.Tests.ps1
```diff
@@ -0,0 +1,395 @@
...
+
+    if($error.Count -eq 0) {
+        Write-Host "found"
+        $resolved[0].IPAddress
+    } else {
+        Write-Host "error"
+        $error[0].CategoryInfo.Category
+    }
+}
+$ResolveDNSLocallyCommand = (${function:Resolve-DNSLocally} -replace "`n|`r", ";")
```
aethyrxeo: Can't we move into body of Resolve-DNS?

Test/Tests/Network/DNS.Tests.ps1
```diff
@@ -0,0 +1,395 @@
...
+            $DNSServerRepo = [DNSServerRepo]::New($MultiNode.NM)
+            [Diagnostics.CodeAnalysis.SuppressMessageAttribute(
+                "PSUseDeclaredVarsMoreThanAssignments",
+                "IPAMRepo",
+                Justification="It's actually used."
+            )]
+            $IPAMRepo = [IPAMRepo]::New($MultiNode.NM)
+
+            Write-Log "Starting Test DNS Server on second test bed..."
+            $DefaultDNSServerDir = "C:\DNS_Server"
```
aethyrxeo: What do you think about creating here functions with calls like this `Start-TestDnsServer -Session $MultiNode.Sessions[1]...` and `Start-TestDnsResolver -Session $MultiNode.Sessions[0]...` so we could reduce the size of the BeforeAll block?

jacekchalupka: Fixed

Test/Tests/Network/DNS.Tests.ps1
jacekchalupka: Because it is as constant in whole script, so I didn't want to declare a function and create a string out of it every time I resolve. I moved this part to "declaration area" at the beginning of a script.

PR #369, opened 2018-10-25
Add "when" condition in test stage to run tests on all builds except …

pmarcinkiewicz1:
…nightly.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -235,7 +235,9 @@ pipeline {
 
         stage('Test') {
             agent { label 'tester' }
-            when { environment name: "DONT_CREATE_TESTBEDS", value: null }
+            when { environment name: "DONT_CREATE_TESTBEDS", value: null;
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -235,7 +235,9 @@ pipeline {
 
         stage('Test') {
             agent { label 'tester' }
-            when { environment name: "DONT_CREATE_TESTBEDS", value: null }
+            when { environment name: "DONT_CREATE_TESTBEDS", value: null;
+                   environment name: "NIGHTLY_BUILD", value: null
```
sodar: Maybe we should add `NIGHTLY_BUILD` to other steps related to tests, such as `Deploy` or `Testenv provisioning`, so that only build stage is run?

Either way, I totally forgot about it, but as you can see, when `DONT_CREATE_TESTBEDS` is set only build stage is run. Maybe we should use it instead?

pmarcinkiewicz1:
recheck no bug

PR #372, opened 2018-11-05
Add branch R6.0-WIP

pmarcinkiewicz1:


ansible/roles/zuul/files/layout.yaml
```diff
@@ -58,7 +60,8 @@ pipelines:
             - ^stable/queens$
             - ^stable/newton$
             - ^stable/ocata$
-            - ^production 
+            - ^production
+            - ^R6.0-WIP
```
sodar: Same as above.

ansible/roles/zuul/files/layout.yaml
```diff
@@ -58,7 +60,8 @@ pipelines:
             - ^stable/queens$
             - ^stable/newton$
             - ^stable/ocata$
-            - ^production 
+            - ^production
```
sodar: I think we should add `$` at the end, so that branch like `production-not-production` are not matched.

PR #374, opened 2018-11-05
Remove github deployment key setup.

m-kostrzewa:
We don't need them since all the repositories are public.


m-kostrzewa:
recheck no bug

PR #381, opened 2018-11-08
Remove netstat invocation at agent's service startup

aethyrxeo:


m-kostrzewa:
Why remove?

aethyrxeo:
Because the bug didn't appear since we have added the netstat invocation

PR #382, opened 2018-11-08
Allow tests to be run on a already prepared environment

mzaremba04:


Test/Tests/Component/NodeManager.Tests.ps1
```diff
@@ -123,20 +124,15 @@ Describe "Node manager" {
...
         Start-NodeMgr -Session $Session
     }
 
     AfterEach {
         try {
             Stop-NodeMgr -Session $Session
-            Clear-TestConfiguration -Session $Session -SystemConfig $MultiNode.Configs.System
         } finally {
-            Merge-Logs -LogSources (New-FileLogSource -Path (Get-ComputeLogsPath) -Sessions $Session)
+            Merge-Logs -DontCleanUp -LogSources (New-FileLogSource -Path (Get-ComputeLogsPath) -Sessions $Session)
```
m-kostrzewa: `-DontCleanUp` will cause subsequent test case log files to contain all the logs from previous test cases I think?

Test/TestConfigurationUtils.ps1
```diff
@@ -319,26 +319,47 @@ function Clear-TestConfiguration {
     Wait-RemoteInterfaceIP -Session $Session -AdapterName $SystemConfig.AdapterName
 }
 
+function Install-ComputeServices {
```
m-kostrzewa: I'm thinking that we shouldn't add anything to TestConfigurationUtils file. If anything, it should shrink. Maybe let's use one of the files in ComputeNode/ directory: https://github.com/Juniper/contrail-windows-ci/tree/master/Test/Utils/ComputeNode

Test/Utils/ComputeNode/Initialize.ps1
```diff
@@ -9,14 +9,45 @@ function Initialize-ComputeNode {
     Param (
         [Parameter(Mandatory=$true)] [PSSessionT] $Session,
         [Parameter(Mandatory=$true)] [TestenvConfigs] $Configs,
-        [Parameter(Mandatory=$true)] [Network[]] $Networks
+        [Parameter(Mandatory=$false)] [bool] $PrepareEnv=$true
```
m-kostrzewa: This goes for `Clear-ComputeNode` as well.

Could you explain why we pass `$PrepareEnv` to this function and then we decide to do nothing? Why not put this logic in clients for this function? E.g. instead of

```
Clear-ComputeNode -Session $Session -SystemConfig $MultiNode.Configs.System -PrepareEnv $PrepareEnv
```

have

```
if($PrepareEnv) { Clear-ComputeNode -Session $Session -SystemConfig $MultiNode.Configs.System }
```

Test/Tests/Component/NodeManager.Tests.ps1
```diff
@@ -1,6 +1,7 @@
 Param (
     [Parameter(Mandatory=$false)] [string] $TestenvConfFile,
     [Parameter(Mandatory=$false)] [string] $LogDir = "pesterLogs",
+    [Parameter(Mandatory=$false)] [bool] $PrepareEnv = $true,
```
sodar: Maybe we should somewhere in the README, that on the setup deployed with CAD, a user should specify `$PrepareEnv` to `$false` before running tests.

Test/Tests/Network/FloatingIP.Tests.ps1
```diff
@@ -89,9 +92,22 @@ Describe "Floating IP" {
...
+                    -Session $Session `
+                    -Networks $Networks `
+                    -Configs $MultiNode.Configs `
+                }
+
             }
 
             AfterAll {
+                foreach ($Session in $MultiNode.Sessions) {
+                    Remove-DockerNetwork -Session $Session -Name $ServerNetwork.Name
```
sodar: Since this PR introduces `$Networks`, we can just iterate over it instead of calling it for each network separately.

Test/Tests/Component/NodeManager.Tests.ps1
mzaremba04: Good point, forgot about it

Test/TestConfigurationUtils.ps1
mzaremba04: Ok, I'd move it to better location.

Test/Utils/ComputeNode/Initialize.ps1
mzaremba04: Well, I thought it would make tests more readable, but I can change if you'd prefer it other way.

Test/Tests/Component/NodeManager.Tests.ps1
mzaremba04: All services would have to be stopped in order to gather logs without DontCleanUp flag and I'd argue that stopping every service after each test case isn't best practice. However, I think I'd like to change it and don't pass prepareenv flag to this tests altogether (i.e. components test) and limit this functionality to network tests only. What do you think?

Test/Utils/ComputeNode/Initialize.ps1
```diff
@@ -2,21 +2,54 @@
...
+        Clear-Logs -LogSources (New-FileLogSource -Path (Get-ComputeLogsPath) -Sessions $Session)
+
+        Write-Log "Uninstalling components from testbed..."
+        Uninstall-Components -Session $Session
+    } else {
+        Write-Log "Not performing environment cleanup because PrepareEnv flag is set to $PrepareEnv"
+    }
+}
 
+function Initialize-DockerNetworks {
```
m-kostrzewa: This function feels like it doesn't belong here. It's kind of test-case specific. Maybe a separate file?

Test/Utils/ComputeNode/Initialize.ps1
```diff
@@ -26,3 +59,44 @@ function Initialize-ComputeNode {
...
+        -OpenStackConfig $OpenStackConfig `
+        -ControllerConfig $ControllerConfig
+}
+
+function Initialize-ComputeServices {
+    Param (
+        [Parameter(Mandatory = $true)] [PSSessionT] $Session,
+        [Parameter(Mandatory = $true)] [SystemConfig] $SystemConfig,
+        [Parameter(Mandatory = $true)] [OpenStackConfig] $OpenStackConfig,
+        [Parameter(Mandatory = $true)] [ControllerConfig] $ControllerConfig
```
m-kostrzewa: 1. 

We have this triple `SystemConfig`, `OpenStackConfig` and `ControllerConfig` passed here all the way from `Initialize-ComputeNode`. Why not just have 
```
[Parameter(Mandatory=$true)] [TestenvConfigs] $Configs
```
here, in `Initialize-ComputeServices`?

2. 

This whole chain of calls Initialize-ComputeNode -> Install-ComputeServices -> Initialize-ComputeServices..

I feel like it should be Initialize-ComputeNode -> Install-ComputeServices && Initialize-ComputeServices

Maybe even something like:

```
function Initialize-ComputeNode {
    Install-ComputeServices
    Set-ComputeServicesConfiguration
    Start-ComputeServices
}

function Install-ComputeServices {
    Install-Components
}

function Set-ComputeServicesConfiguration {
    New-NodeMgrConfigFile
    New-CNMPluginConfigFile
    New-AgentConfigFile
}

function Start-ComputeServices {
    Initialize-DriverAndExtension
    Start-AgentService
}
```

This way we don't create this scary chain of function calls and everything is kinda flat.

3. 

Also, maybe a rename Initialize-ComputeNode -> Initialize-FullWindowsComputeNode

Name changes are really nitpicky and I hate suggesting them and I'm sorry. But I'm always super scared about our init/cleanup code in these tests.

Test/Tests/Component/NodeManager.Tests.ps1
```diff
@@ -148,11 +151,21 @@ Describe "Node manager" {
             "MultiNode",
             Justification="It's actually used."
         )]
-        $MultiNode = New-MultiNodeSetup -TestenvConfFile $TestenvConfFile -InstallNodeMgr
+        $MultiNode = New-MultiNodeSetup -TestenvConfFile $TestenvConfFile
+
+        foreach ($Session in $MultiNode.Sessions) {
+            Install-Components -Session $Session
```
m-kostrzewa: Is it possible to use our brand-new shiny {Initialize,Clear}-ComputeNode in this test?

Test/Utils/ComputeNode/Initialize.ps1
```diff
@@ -2,21 +2,54 @@
...
+        [Parameter(Mandatory = $true)] [SystemConfig] $SystemConfig,
+        [Parameter(Mandatory=$false)] [bool] $PrepareEnv=$true
     )
 
-    Initialize-ComputeServices -Session $Session `
-        -SystemConfig $Configs.System `
-        -OpenStackConfig $Configs.OpenStack `
-        -ControllerConfig $Configs.Controller
+    if ($PrepareEnv) {
+        Clear-TestConfiguration -Session $Session -SystemConfig $SystemConfig
```
m-kostrzewa: While we're at it, can we take a look at `Clear-TestConfiguration`?

It: 

1. cleans up docker networks
2. stops compute node services (but not all!, e.g. doesn't stop node mgr)
3. waits for IP address 

I think we could do something similar like in Initialize-.

```
function Clear-WindowsComputeNode {
    Remove-UnusedDockerNetworks
    Stop-ComputeServices
    Wait-ForIPAddress
    Clear-Logs
    Uninstall-ComputeServices
}
```

The "cleans up docker networks" part we can move to the same location as Initialize-DockerNetworks function maybe?

What I'm suggesting here is proudly presenting what Initialize/Clear logic does in this file, not hiding it in some function in utils somewhere. Of course, we need to keep this file as clear as possible so that it's possible to tell in a glimpse based on function names what exactly will happen.

Test/Tests/Component/NodeManager.Tests.ps1
mzaremba04: Yes. But if we want to invoke merge-logs witout DontCleanup flag then services have to be stopped after every test and started before every test, so it will be redundant to start then in BeforeAll clause.

Test/README.md
```diff
@@ -126,6 +126,15 @@ Run selected test:
 Invoke-Pester -Script @{ Path = ".\TunnellingWithAgent.Tests.ps1"; Parameters = @{ TestenvConfFile = "testenv-conf.yaml"}; } -TestName 'Tunnelling with Agent tests'
 ```
 
+### Running tests on an already deployed environment
+
+Smoke tests can be run on an already deployed environment (for example by contrail-ansible-deployer). Setting PrepareEnv
+flag to false will cause tests to not setup any services and become independant from deployment method:
+
+```
+ Invoke-Pester -Tag 'Smoke' -Script @{ Path = './Tests/Network/*'; Parameters = @{ PrepareEnv = $false; TestenvConfFile="testenv-conf.yaml" };}
```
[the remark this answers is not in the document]
mzaremba04: That would mean that apart from Tags we would also have to pass PrepareEnv flag to Invoke-ProductTests. Don't you think that would clutter it?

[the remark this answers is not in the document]
m-kostrzewa: Or, we can assume that -Smoke flag will append PrepareEnv to Invoke-PesterTests. Or we could not assume that. There are many options here.

How about `Invoke-ProductTests -UsingExistingDeployment`, which would set PrepareEnv and Smoke?

My point is that running smoke tests on existing env seems like a good thing to have a script for, because we want devs to use it easily

Test/Utils/ComputeNode/Initialize.ps1
m-kostrzewa: 2.

you're right https://github.com/Juniper/contrail-windows-ci/blob/6f5171fc82c6a21b66871a04d1ecc0f723dfdfd9/Test/Utils/ComputeNode/Configuration.ps1#L167

meh... I still think having something like the function below is better than 3-level chain, although it's less elegant

```
function Initialize-ComputeNode {
    Install-ComputeServices

    New-NodeMgrConfigFile
    New-CNMPluginConfigFile
    Start-DriverAndExtension

    New-AgentConfigFile
    Start-AgentService
}
```

Test/Tests/Component/NodeManager.Tests.ps1
m-kostrzewa: Something is wrong with the logic here. We do want to have logging per test case, even if services are not restarted between test cases. 

Regarding initialization Initialize-ComputeNode, it seems like it's only useful for smoke tests. Since nodemgr tests aren't smoke, maybe we shouldn't try to force Initialize-ComputeNode usage into them.

m-kostrzewa: http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/2737/TestReports/WindowsCompute/detailed_logs/Tunneling%20with%20Agent%20tests.Tunneling%20MPLSoGRE.ICMP%20-%20Ping%20between%20containers%20on%20separate%20compute%20nodes%20succeeds.txt

http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/2737/TestReports/WindowsCompute/detailed_logs/Tunneling%20with%20Agent%20tests.Tunneling%20MPLSoGRE.IP%20fragmentation%20-%20ICMP%20-%20Ping%20with%20big%20buffer%20succeeds.txt

http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/2737/TestReports/WindowsCompute/detailed_logs/Tunneling%20with%20Agent%20tests.Tunneling%20MPLSoGRE.IP%20fragmentation%20-%20UDP%20-%20sending%20big%20buffer%20succeeds.txt

http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/2737/TestReports/WindowsCompute/detailed_logs/Tunneling%20with%20Agent%20tests.Tunneling%20MPLSoGRE.TCP%20-%20HTTP%20connection%20between%20containers%20on%20separate%20compute%20nodes%20succeeds.txt

http://logs.opencontrail.org/winci/github/WinContrail/contrail-win-ci-gh/2737/TestReports/WindowsCompute/detailed_logs/Tunneling%20with%20Agent%20tests.Tunneling%20MPLSoGRE.UDP%20-%20sending%20message%20between%20containers%20on%20separate%20compute%20nodes%20succeeds.txt

These are logs for separate test cases, but the contents are redundant. The whole point of having these weird loggers was to separate them and I think it's a useful feature. I think we should strive to keep it this way

mzaremba04:
Let's leave it as it is and merge it after changes to logs are completed. I don't feel comfortable with stopping and starting services after every test case as I feel it defies the very idea behind this pull request - being environment independent and safe for the aforementioned environment.

Test/README.md
```diff
@@ -135,6 +136,13 @@ flag to false will cause tests to not setup any services and become independant
  Invoke-Pester -Tag 'Smoke' -Script @{ Path = './Tests/Network/*'; Parameters = @{ PrepareEnv = $false; TestenvConfFile="testenv-conf.yaml" };}
 ```
 
+You can also run it by invoking product tests script:
+
+```
+.\Invoke-ProductTests.ps1 -TestRootDir . -TestenvConfFile ..\testenv-conf.yaml -TestReportDir ../reports -UseExtistingServices
```
m-kostrzewa: UseExtistingServices typo

PR #384, opened 2018-11-09
Removed vrouter utils PDB files from containers

jacekchalupka:


jacekchalupka:
recheck nobug

CIScripts/Build/BuildFunctions.ps1
```diff
@@ -161,7 +160,6 @@ function Invoke-ExtensionBuild {
         if($BuildMode -eq "debug") {
```
michal-clapinski: could you please remove that check? we'd like to have pdbs also in production mode

PR #404, opened 2018-11-29
Simplified services and instances.yaml

michal-clapinski:


Test/Utils/ComputeNode/Service.ps1
```diff
@@ -128,28 +128,23 @@ function New-ServiceConfiguration {
...
         [Parameter(Mandatory=$true)] [string] $ExecutablePath,
         [Parameter(Mandatory=$true)] [string] $LogFileName,
-        [Parameter(Mandatory=$true)] [string[]] [AllowEmptyCollection()] $CommandLineParams
+        [Parameter(Mandatory=$false)] [string[]] $CommandLineParams = @()
     )
     return [Service]::new($ServiceName, $ExecutablePath, $LogFileName, $CommandLineParams)
 }
 
 function Get-AgentServiceConfiguration {
-    $ConfigPath = Get-DefaultAgentConfigPath
```
sodar: `Get-DefaultAgentConfigPath` is not used anywhere anymore, so it can be removed.

michal-clapinski: I removed this field from testenv.yaml instead and started using that function

PR #420, opened 2018-12-06
New Contrail API - Floating IP

jacekchalupka:


Test/Utils/ContrailAPI_New/Models/FloatingIpPool.ps1
```diff
@@ -0,0 +1,25 @@
+class FloatingIpPool : BaseResourceModel {
+    [String] $Name
+    [String] $NetworkName
```
mzaremba04: Why isn't it NetworkFqdn? In FloatingIP there is PoolFqName, why is it different?

Test/Utils/ContrailAPI_New/Models/FloatingIp.ps1
```diff
@@ -0,0 +1,46 @@
...
+        $this.Name = $Name
+        $this.PoolFqName = $PoolFqName
+        $this.Address = $Address
+    }
+
+    [String[]] GetFqName() {
+        return ($this.PoolFqName + @($this.Name))
+    }
+
+    [Hashtable] GetRequest() {
```
mzaremba04: What is the purpose of GetRequest command? Is it used for all of CRUD actions? I think that floating ip address is read-only property, so sending it for update will cause an error.

jacekchalupka: I checked, it does not. You can update the floating_ip_address with in existing floating-ip with request like this.

Test/Utils/ContrailAPI_New/Models/FloatingIpPool.ps1
jacekchalupka: Good point. Fixed. I need to work on my consistency in things like this. Thanks! :)

Test/Utils/ContrailAPI_New/Models/FloatingIp.ps1
mzaremba04: Thanks for checking. It's weird, in schema this field is marked as CR so it misled me. 
Does API server allow changing the address? What if somebody tries to change the IP and then update the floating ip?

jacekchalupka: Yeah, according to docs it is not field that can be passed when using PUT method (updating object). But I just did it, and even in WebUI IP changed and everything just worked fine :)

PR #425, opened 2018-12-10
 Added testbed class to configuration

jacekchalupka:


CIScripts/Testenv/Testenv.ps1
```diff
@@ -22,32 +22,49 @@ class ControllerConfig {
...
 class TestenvConfigs {
-    [SystemConfig] $System;
-    [OpenStackConfig] $OpenStack;
-    [ControllerConfig] $Controller;
+    [SystemConfig] $System
+    [OpenStackConfig] $OpenStack
+    [ControllerConfig] $Controller
+    [TestbedConfig[]] $Testbeds
+
+    TestenvConfigs([String] $TestenvConfFile) {
```
jacekchalupka: I am not sure about this constructor. My thinking:
1) create one TestenvConfigs object in test file,
2) pass it's fields to functions that need config/

Other way would be just to remove this class and create in test 4 objects, one for each config type. I don't want the whole TestenvConfigs object to be passed around.

PR #431, opened 2018-12-17
Changed place where config is read

jacekchalupka:


Test/Tests/Component/NodeManager.Tests.ps1
```diff
@@ -1,27 +1,24 @@
...
 . $PSScriptRoot\..\..\Utils\ComputeNode\Initialize.ps1
 . $PSScriptRoot\..\..\Utils\ComputeNode\Service.ps1
-
 . $PSScriptRoot\..\..\Utils\ComputeNode\Configuration.ps1
-
 . $PSScriptRoot\..\..\TestConfigurationUtils.ps1
 . $PSScriptRoot\..\..\Utils\ContrailNetworkManager.ps1
 . $PSScriptRoot\..\..\Utils\MultiNode\ContrailMultiNodeProvisioning.ps1
 
+
```
aethyrxeo:
```suggestion
```

Test/Tests/Component/AgentRestart.Tests.ps1
```diff
@@ -112,52 +107,56 @@ Test-WithRetries 3 {
         }
 
         BeforeAll {
+            $Testenv = [TestenvConfigs]::New($TestenvConfFile)
             Initialize-PesterLogger -OutDir $LogDir
-            $MultiNode = New-MultiNodeSetup -TestenvConfFile $TestenvConfFile
+            $MultiNode = New-MultiNodeSetup `
+                -Testbeds $Testenv.Testbeds `
+                -ControllerConfig $Testenv.Controller `
```
aethyrxeo: What about passing just $Testenv variable when using more than two fields of the class?

Test/Tests/Component/AgentRestart.Tests.ps1
```diff
@@ -1,19 +1,16 @@
...
+    [Parameter(Mandatory = $false)] [string] $LogDir = "pesterLogs",
+    [Parameter(Mandatory = $false)] [bool] $PrepareEnv = $true,
+    [Parameter(ValueFromRemainingArguments = $true)] $UnusedParams
 )
 
 . $PSScriptRoot\..\..\..\CIScripts\Common\Aliases.ps1
 . $PSScriptRoot\..\..\..\CIScripts\Common\Init.ps1
-
 . $PSScriptRoot\..\..\..\CIScripts\Testenv\Testbed.ps1
-
```
aethyrxeo: IMO the newlines here improve readabillity and i would leave them as they were.

Test/Tests/Component/AgentRestart.Tests.ps1
jacekchalupka: The idea is not to do this :) This will only make code few lines shorter, but the usage will be obscured.

Test/Tests/Component/NodeManager.Tests.ps1
jacekchalupka: fixed

Test/Tests/Component/AgentRestart.Tests.ps1
jacekchalupka: fixed

jacekchalupka:
recheck no bug

PR #433, opened 2018-12-19
Gave responsibilities to Testbed class

jacekchalupka:


CIScripts/Testenv/Testbed.Impl.ps1
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    hidden [PSCredentialT] GetCredentials() {
+        if (-not ($this.Username -or $this.Password)) {
+            return Get-Credential # assume interactive mode
+        }
+        else {
+            $VMUsername = Get-UsernameInWorkgroup -Username $this.Username
+            $VMPassword = $this.Password | ConvertTo-SecureString -AsPlainText -Force
+            return New-Object PSCredentialT($VMUsername, $VMPassword)
```
aethyrxeo: Please use:

```suggestion
            return [PSCredentialT]::new($VMUsername, $VMPassword)
```
It is significantly faster.

CIScripts/Testenv/Testbed.Impl.ps1
```diff
@@ -0,0 +1,73 @@
+class Testbed {
+    [string] $Name
+    [string] $VMName
```
aethyrxeo: What is the difference between these two variables?

jacekchalupka: The second one is used when You have the setup on HyperV and connect to them not by network but some HyperV names.

PR #435, opened 2018-12-19
!!! DO NOT MERGE !!! Testing if flake on fail only

jacekchalupka:


jacekchalupka:
Implemented another approach in other PR.

PR #437, opened 2018-12-20
 Retry 'single compute node' test

jacekchalupka:
This change just surrounded Describe with Test-WithRetries 3 block.


jacekchalupka:
recheck no bug

jacekchalupka:
recheck no bug

PR #442, opened 2019-01-03
Test if VMSwitch exist before starting CNMPlugin

jacekchalupka:


Test/TestConfigurationUtils.ps1
```diff
@@ -267,6 +282,9 @@ function Initialize-CnmPluginAndExtension {
 
     $NRetries = 3;
     foreach ($i in 1..$NRetries) {
+        if(Test-IfVmSwitchExist -Session $Session -VmSwitchName $SystemConfig.VMSwitchName()) {
```
mzaremba04: Is it required? Cnm-plugin can deal with situation where vmswitch already exists as well as with the situation where extension is already enabled.

jacekchalupka:
Changes in CNM-Plugin made this PR not needed.

PR #456, opened 2019-01-08
Jenkinsfile for deployer's template creation

aethyrxeo:


aethyrxeo:
`job/WinContrail-infra/job/templates/job/create-deployer-template/5/`

PR #465, opened 2019-01-11
firewall-rule Contrail API

jablonskim:


Test/Utils/ContrailAPI/Models/FirewallService.ps1
```diff
@@ -0,0 +1,25 @@
+# Those are just informative to show dependencies
+#include "Protocol.ps1"
+#include "PortRange.ps1"
+
+class FirewallService {
+    [Protocol] $Protocol = [Protocol]::any
```
jacekchalupka: Why default value when there is only one constructor that sets $Protocol?

jablonskim:
recheck no bug

jablonskim:
recheck no bug

jablonskim:
recheck no bug

PR #474, opened 2019-01-15
Use custom hashtable for NSSM configuration

michal-clapinski:


Test/Utils/ComputeNode/Service.ps1
```diff
@@ -48,6 +46,14 @@ NSSM output: $($Output.Output)
 "@
         throw [HardError]::new($ExceptionMessage)
     }
+
+    ForEach ($Pair in $AdditionalParams.GetEnumerator()) {
```
aethyrxeo:
```suggestion
    ForEach ($Key in $AdditionalParams.Keys) {
        $Pair = $AdditionalParams[$Key]
```

Don't know why, but GetEnumerator doesn't work sometimes as intented - it returns type instead of value.

michal-clapinski:
recheck no bug

Test/Utils/ComputeNode/Service.ps1
```diff
@@ -6,48 +6,52 @@
...
+        nssm install $Using:Configuration.ServiceName "$Using:Configuration.ExecutablePath"
     } -AllowNonZero -CaptureOutput
 
     $NSSMServiceAlreadyCreatedError = 5
     if ($Output.ExitCode -eq 0) {
         Write-Log $Output.Output
     }
     elseif ($Output.ExitCode -eq $NSSMServiceAlreadyCreatedError) {
-        Write-Log "$ServiceName service already created, continuing..."
+        Write-Log "$Configuration.ServiceName service already created, continuing..."
```
jacekchalupka: Otherwise it will resolve $Configuration and concatenate '.Servicename'
```suggestion
        Write-Log "$($Configuration.ServiceName) service already created, continuing..."
```

Test/Utils/ComputeNode/Service.ps1
```diff
@@ -210,52 +157,62 @@ function Test-IsCNMPluginServiceRunning {
...
 }
 
 function New-CNMPluginService {
     Param (
         [Parameter(Mandatory=$true)] $Session
     )
 
-    New-RemoteService `
+    $AdditionalParams = @{
+        AppStdout = (Get-CNMPluginServiceLogPath)
```
aethyrxeo: Embrace key values which are strings with single quotes.

PR #476, opened 2019-01-17
Testing for flakiness in detailed logs

jacekchalupka:


flakes/grep-dir.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+set -u
+set -o pipefail
+
+IFS=$'\n';
```
mzaremba04: What is it for?

jacekchalupka: It is variable that sets how to split input in foreach. By default it is set to any whitespace, and in that scenario filenames with spaces in it are splited into few iterations of loop. When I set it to \n, whole lines are passed to each iteration.

jacekchalupka:
recheck no bug

PR #493, opened 2019-01-24
Freezing version of psutil to 5.4.8

jacekchalupka:


jacekchalupka:
recheck no bug

jacekchalupka:
recheck no bug

PR #502, opened 2019-01-31
Improve test initialization

jacekchalupka:
All functions used during initialization are transactional.
Improved initialization logging.


jacekchalupka:
recheck no bug

PR #505, opened 2019-02-01
 Checking vswitch state before waiting for IP

jacekchalupka:


Test/TestConfigurationUtils.ps1
```diff
@@ -197,6 +197,45 @@ function Wait-RemoteInterfaceIP {
     } | Out-Null
 }
 
+function Test-IfVmSwitchExist {
+    Param ([Parameter(Mandatory = $true)] [PSSessionT] $Session,
```
michal-clapinski: pls split into multiple lines like in `Assert-VmSwitchInitialized`

Test/TestConfigurationUtils.ps1
```diff
@@ -197,6 +197,45 @@ function Wait-RemoteInterfaceIP {
...
+        [Parameter(Mandatory = $true)] [String] $VmSwitchName)
+
+     $r = Invoke-Command -Session $Session -ScriptBlock {
+        Get-VMSwitch $Using:VMSwitchName -ErrorAction SilentlyContinue | Select-Object -ExpandProperty isDeleted
+    }
+
+     if(($null -eq $r) -or ($r -eq $true)) {
```
michal-clapinski: What about some .equals() method instead?

Test/TestConfigurationUtils.ps1
```diff
@@ -197,6 +197,45 @@ function Wait-RemoteInterfaceIP {
...
+        return $false
+    }
+
+     return $true
+}
+
+function Assert-VmSwitchInitialized {
+    Param (
+        [Parameter(Mandatory = $true)] [PSSessionT] $Session,
+        [Parameter(Mandatory = $true)] [SystemConfig] $SystemConfig
```
michal-clapinski: Are you sure you want to use `SystemConfig` instead of the two fields?

Test/TestConfigurationUtils.ps1
jacekchalupka: Fixed

Test/TestConfigurationUtils.ps1
jacekchalupka: Fixed

Test/TestConfigurationUtils.ps1
jacekchalupka: Yes. That way usage of both asserts is the same, and more important there is can be no bug in usage, for example user calls Assert-VmSwitchDeleted and waits on virtual interface.

PR #509, opened 2019-02-05
Test script run in jenkins exit 0 when tests pass

jacekchalupka:
Without exit 0 at the end of script jenkins will fail if something
has been written in error stream. Exceptions will still cause job
to fail, because they will prevent exit 0 from being executed.


jacekchalupka:
recheck no bug

PR #525, opened 2019-02-22
Add new env vars that'll be used by scons soon

michal-clapinski:
Needed for: https://review.opencontrail.org/#/c/49677/


michal-clapinski:
recheck no bug

PR #527, opened 2019-02-28
Dockerfile to build containers parametrized

jacekchalupka:
For Windows Server 2019 we will need to use other base images for
containers, that's why we need to parametrize Dockerfile.

Also on WS 2019, nanoserver doesn't have powershell, so we cannot
unpack zip with artifacts.


jacekchalupka:
recheck no bug

PR #535, opened 2019-03-08
Fixing controller to use two interfaces

jacekchalupka:


jacekchalupka:
recheck no bug

jacekchalupka:
recheck no bug

jacekchalupka:
recheck no bug

PR #536, opened 2019-03-11
Stop accepting tests with non-zero exit codes

michal-clapinski:


michal-clapinski:
recheck no bug

michal-clapinski:
recheck no bug

michal-clapinski:
recheck no bug

PR #540, opened 2019-03-14
Add data adapter ip info to testbed class

mzaremba04:


Test/Utils/Testenv/Testenv.Impl.ps1
```diff
@@ -49,6 +49,7 @@ class Testenv {
 
         if ($InstallComponents) {
             foreach ($Testbed in $this.Testbeds) {
+                $Testbed.SetDataAdapterIpInfo()
```
jacekchalupka: Why don't we do it in constructor?

mzaremba04: It's only neede if components are bing installed


There are only two places where this info is needed in clear test configuration and in initialize cnm plugin and extension. This is strictly used when components are being installed. I can do it both ways but it won't be used for anything.

Test/Utils/Testenv/Testbed.Impl.ps1
```diff
@@ -13,6 +13,10 @@ class Testbed {
     [String] $VHostName
     [PSSessionT] $Session = $null
 
+    # Keep in mind that this vairiable will only be initialized when data adapter exists at the beginning of run
```
jacekchalupka: I believe we don't need this comment anymore

Test/Tests/Component/VTestScenarios.Tests.ps1
```diff
@@ -25,6 +25,7 @@ Describe 'vTest scenarios' -Tag Smoke {
 
     BeforeAll {
         $Testbeds = [Testbed]::LoadFromFile($TestenvConfFile)
+        $Testbeds[0].SetDataAdapterIpInfo()
```
jacekchalupka: This line is not needed, since Testbed constructor does that.

Test/Utils/Testenv/Testbed.Impl.ps1
```diff
@@ -149,6 +154,29 @@ class Testbed {
...
+                    return $Res
+                }
+                return @{
+                    IPAddress = $Res.IPAddress;
+                    PrefixLength = $Res.PrefixLength;
+                }
+            }
+        }
+
+        $this.DataIpInfo = Get-IpInfo($this.DataAdapterName)
```
jacekchalupka: Get-IpInfo is a function, so it need to be called like this:
`$this.DataIpInfo = Get-IpInfo -adapter $this.DataAdapterName`
But maybe just make it a method that can be used in other places of code? I have in mind at least one where we need to get IP from specific adapter from testbed.

mzaremba04: Thanks, missed that.
About extracting this part of code: I think it should be done when it's needed not in anticiptation for future problems.

Test/Utils/Testenv/Testbed.Impl.ps1
mzaremba04: Thanks

jacekchalupka:
recheck no bug

jacekchalupka:
recheck no bug

PR #547, opened 2019-03-20
Fixing Nodemanager test to work on WS2019

jacekchalupka:


Test/Tests/Component/NodeManager.Tests.ps1
```diff
@@ -22,23 +22,8 @@ Param (
 . $PSScriptRoot\..\..\Utils\MultiNode\ContrailMultiNodeProvisioning.ps1
 
 $ContrailProject = 'ci_tests_nodemanager'
-
-# TODO: This variable is not passed down to New-NodeMgrConfig in ComponentsInstallation.ps1
-#       Should be refactored.
 $LogPath = Get-NodeMgrLogPath
```
michal-clapinski: Since it's currently used only in one place, could you move this variable to `Test-NodeMgrLogs` function?

PR #556, opened 2019-03-21
Vrouter certificate is now imported in test phase

jacekchalupka:
vRouter is no longer signed with codilime cert, but with one that is
generated by Visual Studio and exported to vRouter.cer file.

### DIFF
--- a/CIScripts/Build.ps1
+++ b/CIScripts/Build.ps1
@@ -42,16 +42,9 @@ foreach ($Directory in $Directories) {
 
 try {
     Invoke-CnmPluginBuild -PluginSrcPath $Env:CNM_PLUGIN_SRC_PATH `
-        -SigntoolPath $Env:SIGNTOOL_PATH `
-        -CertPath $Env:CERT_PATH `
-        -CertPasswordFilePath $Env:CERT_PASSWORD_FILE_PATH `
-        -OutputPath $CnmPluginOutputDir `
-        -LogsPath $LogsDir
+        -OutputPath $CnmPluginOutputDir
 
     Invoke-ExtensionBuild -ThirdPartyCache $Env:THIRD_PARTY_CACHE_PATH `
-        -SigntoolPath $Env:SIGNTOOL_PATH `
-        -CertPath $Env:CERT_PATH `
-        -CertPasswordFilePath $Env:CERT_PASSWORD_FILE_PATH `
         -BuildMode $SconsBuildMode `
         -OutputPath $vRouterOutputDir `
         -LogsPath $LogsDir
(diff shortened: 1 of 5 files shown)


CIScripts/Build.ps1
```diff
@@ -42,26 +42,16 @@ foreach ($Directory in $Directories) {
...
-        -LogsPath $LogsDir
+        -OutputPath $CnmPluginOutputDir
 
     Invoke-ExtensionBuild -ThirdPartyCache $Env:THIRD_PARTY_CACHE_PATH `
-        -SigntoolPath $Env:SIGNTOOL_PATH `
-        -CertPath $Env:CERT_PATH `
-        -CertPasswordFilePath $Env:CERT_PASSWORD_FILE_PATH `
         -BuildMode $SconsBuildMode `
         -OutputPath $vRouterOutputDir `
         -LogsPath $LogsDir
```
michal-clapinski: Are those even useful anymore? You could remove those too :D

Test/Utils/ComputeNode/Installation.ps1
```diff
@@ -62,6 +62,34 @@ function Uninstall-Agent {
...
 }
 
+function Import-ExtensionCertificate {
+    Param ([Parameter(Mandatory = $true)] [Testbed] $Testbed)
+
+    Write-Log 'Importing vRouter certificate to certstore'
+    Invoke-Command -Session $Testbed.GetSession() {
+        $Cert = Import-Certificate -CertStoreLocation Cert:\LocalMachine\Root\ 'C:\Artifacts\vrouter\vRouter.cer'
+        try {
+            Import-Certificate -CertStoreLocation Cert:\LocalMachine\TrustedPublisher\ 'C:\Artifacts\vrouter\vRouter.cer'
```
michal-clapinski: why is one in `try` block but the other one is not? please add comment to the code about this

jacekchalupka: Isn't the catch block explaining why?

CIScripts/Build.ps1
jacekchalupka: You mean logs?

michal-clapinski: Yeah. I went to the logs folder and couldn't find `vrouter_build.log` for example. And even if it's there, it's rather useless, right?

Test/Utils/ComputeNode/Installation.ps1
jacekchalupka: Yes, that is the reason. Importing twice will not break anything, but the idea behind our tests is: cleanup after yourself if it is possible, leave the place as you found it. That's why all steps that we run during initialization should be atomic in a sense, that they either success or fail, but leave environment untouched. If the function is not working like this, it has CleanupStack as a parameter, and after every step it adds cleanup step.

CIScripts/Build.ps1
jacekchalupka: I have no idea why it is here, and I believe I would have to look through code what happens with logs, I don't feel like deleting them without it :) But I will gladly do it tomorrow :)

michal-clapinski: I'll do it. Already working on it while writing proper test runner

jacekchalupka:
recheck no bug

PR #560, opened 2019-03-25
Use Test directory from contrail-windows-test

jacek-izykowski:


jacek-izykowski:
recheck no bug

jablonskim:
recheck no bug

PR #564, opened 2019-03-27
WS2019 containers now have tag prefixed winsrv2019

jacekchalupka:


jacekchalupka:
Not needed, because zuulv3 now set the whole tag

PR #557, opened 2019-03-22
Implement proper test runner using scons

michal-clapinski:

PR #568, opened 2019-03-28
Moved running kernel tests to a place with edited path

michal-clapinski:
New cmocka is linked dynamically so I need `build/bin` added to path before running kernel tests. This is just a temporal change, as this code is totally refactored in #557

### DIFF
--- a/CIScripts/Build/BuildFunctions.ps1
+++ b/CIScripts/Build/BuildFunctions.ps1
@@ -90,12 +90,6 @@ function Invoke-ExtensionBuild {
         }
     })
 
-    $Job.Step("Running kernel unit tests", {
-        Invoke-NativeCommand -ScriptBlock {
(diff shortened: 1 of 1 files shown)


CIScripts/Build/BuildFunctions.ps1
```diff
@@ -294,6 +288,10 @@ function Invoke-ProductUnitTests {
             "timeout's threshold. They were copied from Linux unit test job.")]
         $Env:TASK_UTIL_RETRY_COUNT = 6000
 
+        Invoke-NativeCommand -ScriptBlock {
+            scons --opt=$BuildMode kernel-tests vrouter:test
+        }
```
jacekchalupka: Why not Out-Null'ed?

michal-clapinski: 1. Because cut-pasted from the old code.
2. Because I don't care since it'll be removed in #557

PR #572, opened 2019-04-03
Switched to fetch_packages in CI

michal-clapinski:


michal-clapinski:
recheck no bug

michal-clapinski:
recheck no bug

PR #579, opened 2019-05-14
Use fetch_packages.py in CI

michal-clapinski:


michal-clapinski:
recheck no bug

jablonskim:
recheck no bug

